PR #24, opened 2024-08-02
chore: rename config `ChainName` to `ChainType`

bap2pecs:
## Summary
when I am doing the devnet integration, I realized that `ChainName` can cause confusion. e.g. `OPStackL2` is not really a consumer chain name, it's actually a type

after the changes, the config will looks like this
```
[Application Options]
; Logging level for all subsystems
LogLevel = info

; the type of the consumer chain
ChainType = babylon
```

there are only 3 possible types:
```
const (
	BabylonConsumerChainType   = "babylon"
	OPStackL2ConsumerChainType = "OPStackL2"
	WasmConsumerChainType      = "wasm"
)
```

## Test Plan

```
make build
./build/fpd init --home ./.fpd
```
and see the generated config file

run tests
```
make test
make test-e2e-op
make test-e2e-babylon
```


clientcontroller/factory.go
```diff
@@ -13,9 +13,9 @@ import (
 )
 
 const (
-	BabylonConsumerChainName   = "babylon"
-	OPStackL2ConsumerChainName = "OPStackL2"
-	WasmConsumerChainName      = "wasm"
+	BabylonConsumerChainType   = "babylon"
+	OPStackL2ConsumerChainType = "OPStackL2"
+	WasmConsumerChainType      = "wasm"
```
maurolacy, line 18: Why not use an enum / constants here?

maurolacy (approved):

clientcontroller/factory.go
bap2pecs: what do you mean? those enums are already used in code

finality-provider/config/config.go
```diff
@@ -55,7 +55,7 @@ var (
 // Config is the main config for the fpd cli command
 type Config struct {
 	LogLevel                 string        `long:"loglevel" description:"Logging level for all subsystems" choice:"trace" choice:"debug" choice:"info" choice:"warn" choice:"error" choice:"fatal"`
-	ChainName                string        `long:"chainname" description:"the name of the consumer chain"`
+	ChainType                string        `long:"chaintype" description:"the type of the consumer chain"`
```
SebastianElvis, line 58: Maybe mention all possible values to the `description`?

SebastianElvis (approved):

clientcontroller/factory.go
maurolacy: Oh, I see. Those _already are_ enums in Go.

Perhaps matching against the const values instead of the strings then?

PR #26, opened 2024-08-05
test: op e2e finality gadget daemon itest

parketh:
## Summary

This PR adds integration tests for the new finality gadget daemon in `babylonlabs-io/finality-gadget`, a stateful program used to track consecutive quorum and query the latest BTC-finalized block.

- feature overview: https://github.com/babylonlabs-io/pm-integration/issues/14
- implementation: https://github.com/babylonlabs-io/finality-gadget/pull/13

## Test Plan
```
make lint
make test-e2e-op
```


SebastianElvis (approved):
:+1:

parketh:
now fixed and ready for review again @SebastianElvis @bap2pecs

SebastianElvis (approved):
Already approved, lgtm :+1:

finality-provider/config/opstackl2.go
```diff
@@ -11,8 +11,9 @@ import (
 )
 
 type OPStackL2Config struct {
-	OPStackL2RPCAddress     string `long:"opstackl2-rpc-address" description:"the rpc address of the op-stack-l2 node to connect to"`
-	OPFinalityGadgetAddress string `long:"op-finality-gadget" description:"the contract address of the op-finality-gadget"`
+	OPStackL2RPCAddress      string `long:"opstackl2-rpc-address" description:"the rpc address of the op-stack-l2 node to connect to"`
+	OPFinalityGadgetAddress  string `long:"op-finality-gadget" description:"the contract address of the op-finality-gadget"`
```
bap2pecs, line 15: iiuc after we use the gadget RPC, we don't need this contract address anymore. wdyt?

note: removing it can be in a separate PR

itest/opstackl2/op_e2e_test.go
```diff
@@ -185,3 +185,58 @@ func TestFinalityStuckAndRecover(t *testing.T) {
...
+	}()
+
+	// check latest block
+	require.Eventually(t, func() bool {
+		block, err := ctm.FinalityGadget.QueryLatestFinalizedBlock()
+		if block == nil {
+			return false
+		}
+		require.NoError(t, err)
+		return block.BlockHeight > targetBlockHeight+6
```
bap2pecs, line 237: does it have to be 6 or it can be a random number? (if it's random, let's add a comment above to document that)

itest/opstackl2/op_test_manager.go
```diff
@@ -52,16 +56,21 @@ type BaseTestManager = base_test_manager.BaseTestManager
...
+	BaseDir              string
+	FinalityGadget       *finalitygadget.FinalityGadget
+	FinalityGadgetClient *fgclient.FinalityGadgetGrpcClient
+	Db                   db.IDatabaseHandler
+	DbPath               string
+	OpSystem             *ope2e.System
 }
 
 func StartOpL2ConsumerManager(t *testing.T, numOfConsumerFPs uint8) *OpL2ConsumerTestManager {
+	defaultlog.SetOutput(os.Stdout)
```
bap2pecs, line 72: just curious, why do we need this?

itest/opstackl2/op_test_manager.go
```diff
@@ -109,18 +124,58 @@ func StartOpL2ConsumerManager(t *testing.T, numOfConsumerFPs uint8) *OpL2Consume
 	require.NoError(t, err)
 	t.Logf(log.Prefix("Register consumer %s to Babylon"), opConsumerId)
 
+	// define finality gadget configs
+	cfg := fgcfg.Config{
+		L2RPCHost:         opL2ConsumerConfig.OPStackL2RPCAddress,
+		BitcoinRPCHost:    "rpc.ankr.com/btc",
```
bap2pecs, line 130: iirc, this is not used. if that's the case, let's use a mock URL to avoid confusion?

itest/opstackl2/op_test_manager.go
```diff
@@ -109,18 +124,58 @@ func StartOpL2ConsumerManager(t *testing.T, numOfConsumerFPs uint8) *OpL2Consume
...
+		FGContractAddress: opL2ConsumerConfig.OPFinalityGadgetAddress,
+		BBNChainID:        "chain-test",
```
bap2pecs, line 132: probably better to use `e2eutils.ChainID`

itest/opstackl2/op_test_manager.go
```diff
@@ -394,24 +451,6 @@ func startEotsManagers(
...
 
-func initSdkClient(
-	opSys *ope2e.System,
-	opL2ConsumerConfig *fpcfg.OPStackL2Config,
-	t *testing.T,
-) *sdkclient.SdkClient {
-	// We pass in an external Bitcoin RPC address but otherwise use the default configs.
-	btcConfig := btcclient.DefaultBTCConfig()
-	// The RPC url must be trimmed to remove the http:// or https:// prefix.
-	btcConfig.RPCHost = trimLeadingHttp(opSys.Cfg.DeployConfig.BabylonFinalityGadgetBitcoinRpc)
```
bap2pecs, line 405: is `trimLeadingHttp` still being used after removing this function? if not let's remove it

itest/opstackl2/op_test_manager.go
```diff
@@ -801,6 +837,10 @@ func (ctm *OpL2ConsumerTestManager) WaitForBlockFinalized(
 	require.Eventually(t, func() bool {
 		// doesn't matter which FP we use to query. so we use the first consumer FP
 		latestFinalizedBlock, err := ctm.getOpCCAtIndex(0).QueryLatestFinalizedBlock()
+		if err != nil {
+			t.Logf(log.Prefix("failed to query latest finalized block %s"), err.Error())
+			return false
+		}
 		require.NoError(t, err)
```
bap2pecs, line 844: let's remove this line since we already checked err != nil above

bap2pecs (approved):
Olympic Games Sport - h4bWxVCqEIn7BxBKKd

finality-provider/config/opstackl2.go
parketh: given `OPFinalityGadgetAddress` is still needed by the finality gadget server, i think it would be better to keep it here and avoid having to store it elsewhere + update the function signatures

itest/opstackl2/op_test_manager.go
parketh: why is it not used? i think we still use it to query bitcoin block timestamps in `finality-gadget`

itest/opstackl2/op_e2e_test.go
parketh: will add a comment to explain

itest/opstackl2/op_test_manager.go
parketh: will delete - was playing around with different logging settings previously and forgot to revert

finality-provider/config/opstackl2.go
bap2pecs: sure we can keep it for now. I was mostly thinking about that FG for different consumer chains will have diff RPC URLs and each consumer chain will have its own gadget address, which will be configured when starting the FP server. wdyt?

itest/opstackl2/op_test_manager.go
bap2pecs: oh so does that mean our e2e test depend on this ankr RPC? maybe I got confused with what we have in the OP unit tests where we used mock url

PR #27, opened 2024-08-07
minor: fix Makefile

SebastianElvis:


SebastianElvis:
OK looks like we haven't rebased FP yet. Repruposing this PR for rebasing then. Converting to draft for now

SebastianElvis:
Merging to unblock the new round of Euphrates devnet. From my side all unit tests and e2e tests are passing. Leftovers:

- OP e2e test needs to be fined (https://github.com/babylonlabs-io/finality-provider/pull/26 seems relevant) cc @bap2pecs @parketh @lesterli 
- GH action needs to be configured such that it can pull private repositories cc @filippos47 could we get any help on this?

Issue #8, opened 2024-08-01
Simulation testing to simulate negative cases


gitferry:
To be sure core logic handles those gracefully

PR #40, opened 2024-08-20
feat: accormodate BTC-timestamping

gitferry:


finality-provider/config/config.go
```diff
@@ -25,8 +25,8 @@ const (
 	defaultFinalityProviderKeyName = "finality-provider"
 	DefaultRPCPort                 = 12581
 	defaultConfigFileName          = "fpd.conf"
-	defaultNumPubRand              = 100
-	defaultNumPubRandMax           = 200
+	defaultNumPubRand              = 1000
+	defaultNumPubRandMax           = 1000
```
SebastianElvis, line 29: Is this enough? I thought we need to commit pub rand that can last for 1-2 weeks

SebastianElvis (commented):
Code wise lgtm, it seems that requiring a FP to have timestamped pubrand for having voting power actually simplies the protocol, which is good :+1: 

One question is that should we add some code for handling/testing the edge case where FP somehow used all timestamped pub rand (eg when BTC gets highly congested). When this happens at height `h`, FP needs to wait until a later height `h'` where it has positive voting power, and then start voting at height `h'`, without voting for `h` to `h'`. This seems not handled in the current PR

gitferry:
> One question is that should we add some code for handling/testing the edge case where FP somehow used all timestamped pub rand (eg when BTC gets highly congested). When this happens at height h, FP needs to wait until a later height h' where it has positive voting power, and then start voting at height h', without voting for h to h'. This seems not handled in the current PR

I think after we connect randomness commit with voting power, the two threads of sending finality signatures and [sending](https://github.com/babylonlabs-io/finality-provider/blob/699da4fd0e29169d0bdf0955f28f0e5c296a9322/finality-provider/service/fp_instance.go#L126) pub rand commit can be independent, i.e., the voting thread does not need to know whether the pub rand commit is there. So I think the current code already covers this case.

Nevertheless, I agree that we should do a major refactoring and add comprehensive tests after the feature is shipped, which is planned in #6 and #8

finality-provider/config/config.go
gitferry: You are right. We have calculated the number as 7k in https://github.com/babylonlabs-io/pm/blob/main/adr/adr-024-timestamping-public-randomness.md

SebastianElvis (approved):
Right, now I remember -- if FP finds itself does not have voting power at this height it will simply skip this block. All good then :+1:

finality-provider/service/fp_instance.go
```diff
@@ -816,14 +711,14 @@ func (fp *FinalityProviderInstance) SubmitBatchFinalitySignatures(blocks []*type
...
-			lastCommittedHeight, b.Height)
-	}
+	// lastCommittedHeight, err := fp.GetLastCommittedHeight()
+	// if err != nil {
+	// 	return nil, nil, err
+	// }
+	// if lastCommittedHeight < b.Height {
+	// 	return nil, nil, fmt.Errorf("the finality-provider's last committed height %v is lower than the current block height %v",
+	// 		lastCommittedHeight, b.Height)
+	// }
```
RafilxTenfen, line 721: can we remove this?

finality-provider/service/fastsync.go
```diff
@@ -62,14 +62,6 @@ func (fp *FinalityProviderInstance) FastSync(startHeight, endHeight uint64) (*Fa
 				fp.metrics.IncrementFpTotalBlocksWithoutVotingPower(fp.GetBtcPkHex())
 				continue
 			}
-			// check whether the randomness has been committed
```
RafilxTenfen, line 65: question: we can stop checking the randomness has been committed, because we don't need to send randomness for past blocks?

RafilxTenfen (approved):
LGTM :rocket:

finality-provider/service/fastsync.go
gitferry: This is because in [ADR-024](https://github.com/babylonlabs-io/pm/blob/main/adr/adr-024-timestamping-public-randomness.md), a fp can only have voting power if it has BTC-timestamped pub rand for the height. At Line 57, we have already checked whether the fp has voting power. This indicates whether the fp has timestamped pub rand or not, so we don't need to check randomness here

PR #42, opened 2024-08-25
chore: update query activated height

lesterli:
## Summary

This PR updated the implementation of activated height which queries the BTC staking activated timestamp via the finality gadget grpc, and then converted it to the L2 block number.

Content https://github.com/babylonchain/finality-provider/issues/432

## Test Plan

```
make lint
make test
make test-e2e-op
```

### DIFF
--- a/clientcontroller/opstackl2/consumer.go
+++ b/clientcontroller/opstackl2/consumer.go
@@ -459,7 +459,7 @@ func (cc *OPStackL2ConsumerController) GetBlockNumberByTimestamp(ctx context.Con
 	}
 
 	// binary search between block 1 and the latest block
-	// start from block 1, b/c calling the HeaderByNumber of block 0, its timestamp is 0
+	// start from block 1, b/c some L2s such as OP mainnet, block 0 is genesis block with timestamp 0
 	lowerBound := uint64(1)
 	upperBound := latestBlock.Number.Uint64()
 
--- a/clientcontroller/opstackl2/consumer.go
+++ b/clientcontroller/opstackl2/consumer.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"math"
 	"math/big"
 
 	sdkErr "cosmossdk.io/errors"
@@ -358,19 +359,19 @@ func (cc *OPStackL2ConsumerController) QueryActivatedHeight() (uint64, error) {
 	finalityGadgetClient, err := fgclient.NewFinalityGadgetGrpcClient(cc.Cfg.BabylonFinalityGadgetRpc)
 	if err != nil {
 		cc.logger.Error("failed to initialize Babylon Finality Gadget Grpc client", zap.Error(err))
-		return 0, err
+		return math.MaxUint64, err
 	}
 
 	activatedTimestamp, err := finalityGadgetClient.QueryBtcStakingActivatedTimestamp()
 	if err != nil {
 		cc.logger.Error("failed to query BTC staking activate timestamp", zap.Error(err))
-		return 0, err
+		return math.MaxUint64, err
 	}
 
 	l2BlockNumber, err := cc.GetBlockNumberByTimestamp(context.Background(), activatedTimestamp)
 	if err != nil {
 		cc.logger.Error("failed to convert L2 block number from the given BTC staking activation timestamp", zap.Error(err))
-		return 0, err
+		return math.MaxUint64, err
 	}
 
 	return l2BlockNumber, nil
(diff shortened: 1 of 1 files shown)


clientcontroller/opstackl2/consumer.go
```diff
@@ -353,25 +354,26 @@ func (cc *OPStackL2ConsumerController) QueryIsBlockFinalized(height uint64) (boo
...
 // QueryActivatedHeight returns the L2 block number at which the finality gadget is activated.
-// It is fetched from the configuration of a CosmWasm contract OP finality gadget.
 func (cc *OPStackL2ConsumerController) QueryActivatedHeight() (uint64, error) {
-	queryMsg := &QueryMsg{Config: &Config{}}
-	jsonData, err := json.Marshal(queryMsg) // `{"config":{}}`
+	finalityGadgetClient, err := fgclient.NewFinalityGadgetGrpcClient(cc.Cfg.BabylonFinalityGadgetRpc)
 	if err != nil {
-		return 0, fmt.Errorf("failed marshaling to JSON: %w", err)
+		cc.logger.Error("failed to initialize Babylon Finality Gadget Grpc client", zap.Error(err))
+		return 0, err
```
bap2pecs, line 361: not sure why we put 0 there in the past. but I feel we should return max of uint64 on error. wdyt? same for the rest place as well as in `GetBlockNumberByTimestamp`

clientcontroller/opstackl2/consumer.go
```diff
@@ -431,6 +433,48 @@ func ConvertProof(cmtProof cmtcrypto.Proof) Proof {
...
+// GetBlockNumberByTimestamp returns the L2 block number for the given BTC staking activation timestamp.
+// It uses a binary search to find the block number.
+func (cc *OPStackL2ConsumerController) GetBlockNumberByTimestamp(ctx context.Context, targetTimestamp uint64) (uint64, error) {
+	// Check if the target timestamp is after the latest block
+	latestBlock, err := cc.opl2Client.HeaderByNumber(ctx, nil)
+	if err != nil {
+		return 0, err
+	}
+	if targetTimestamp > latestBlock.Time {
+		return 0, nil
```
bap2pecs, line 445: this logic seems incorrect. if the target timestamp is in the future, we will be returning 0 as the activated height. also, I feel we should throw an error instead of nil b/c error I think this is a case where we don't expect it to happen. or am I missing anything?

clientcontroller/opstackl2/consumer.go
```diff
@@ -431,6 +433,48 @@ func ConvertProof(cmtProof cmtcrypto.Proof) Proof {
...
+	}
+	// Check if the target timestamp is before the first block
+	firstBlock, err := cc.opl2Client.HeaderByNumber(ctx, big.NewInt(1))
+	if err != nil {
+		return 0, err
+	}
+	if targetTimestamp < firstBlock.Time {
```
bap2pecs, line 452: let's say block 0 is at t0 and block 1 at t1

if `t0 < t < t1`, the activated height should be block 1 but not block 0. so here should return 1 instead. wdyt?

clientcontroller/opstackl2/consumer.go
```diff
@@ -431,6 +433,48 @@ func ConvertProof(cmtProof cmtcrypto.Proof) Proof {
...
+		return 0, nil
+	}
+
+	lowerBound := uint64(1)
```
bap2pecs, line 456: i am also curious, why not compare with block 0 directly but pick block 1?

lesterli: b/c calling the `HeaderByNumber` of block 0, its timestamp is 0

```
curl -X POST https://rpc.ankr.com/optimism -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x0", false],"id":1}'

{"jsonrpc":"2.0","id":1,"result":{"hash":"0x7ca38a1916c42007829c55e69d3e9a73265554b586a499015373241b8a3fa48b","number":"0x0","timestamp":"0x0","transactions":[]}}
```

clientcontroller/opstackl2/consumer.go
lesterli: ok, we can return max of uint64 as the activated height and an error.

clientcontroller/opstackl2/consumer.go
lesterli: yes, it should return 1

clientcontroller/opstackl2/consumer.go
lesterli: yes, max of uint64 makes sense.

clientcontroller/opstackl2/consumer.go
SebastianElvis: Would be good to clarify this in the docstring

SebastianElvis (approved):

bap2pecs (approved):

PR #44, opened 2024-08-28
Feat: impl QueryFinalityProviderHasPower in consumer chain

bournezjc:
# Summary
context in [#496](https://github.com/babylonchain/finality-provider/issues/496)
This pr implements the logic of querying voting power in Op. 

> instead of querying the exact voting power, the FP queries BTC delegations page by page, and if there is >=1 active BTC delegation, then it can vote.

# Test plan
```
make test-e2e-op
```

### DIFF
--- a/clientcontroller/opstackl2/consumer.go
+++ b/clientcontroller/opstackl2/consumer.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"math"
 	bbnclient "github.com/babylonlabs-io/babylon/client/client"
+	btcstakingtypes "github.com/babylonlabs-io/babylon/x/btcstaking/types"
+	sdkquerytypes "github.com/cosmos/cosmos-sdk/types/query"
 	"math/big"
 
 	sdkErr "cosmossdk.io/errors"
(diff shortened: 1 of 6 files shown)


clientcontroller/opstackl2/consumer.go
```diff
@@ -486,3 +542,32 @@ func (cc *OPStackL2ConsumerController) Close() error {
...
+	btcDel *btcstakingtypes.BTCDelegationResponse,
+) (bool, error) {
+
+	btcstakingParams, err := cc.bbnClient.QueryClient.BTCStakingParams()
+	if err != nil {
+		return false, err
+	}
+
+	covQuorum := btcstakingParams.GetParams().CovenantQuorum
+	ud := btcDel.UndelegationResponse
```
SebastianElvis, line 556: Since this never changes and the `isDelegationActive` function will be called upon every BTC delegation, how about we pass `btcstakingParams` as an argument to `isDelegationActive` rather than querying it everytime?

SebastianElvis (approved):

lesterli (approved):

PR #50, opened 2024-09-11
chore: add retry for consumer chain to become available

RafilxTenfen:
The goal here was to make life easier for finality provider operators to be able to turn on the finality provider before the consumer node is available and wait for that node to become available for `MinutesToWaitForConsumer` minutes


Lazar955 (approved):
Nice work 👍

PR #51, opened 2024-09-11
chore: opt to set eots pk in `fpd create-finality-provider`

RafilxTenfen:
New flag to command `create-finality-provider` to allow specify eots-pk as hex in the parameters to avoid creation of new EOTS key to store in database

This will be useful for finality providers that are in Phase-1 and will need to transition to Phase-2. 
During Phase-1, only EOTS keys were generated for Phase-2 these finality providers will need to use the same EOTS pk to load and start their finality provider


Lazar955 (approved):
lgtm 🙌

gitferry (approved):
Great work!

Issue #34, opened 2024-08-12
Remove support for multiple FP instances


SebastianElvis:
PoP verification will fail if fpName is not same as in config. This is because

- upon createFinalityProviderRequest it will create a new secret key in the key ring. The user specifies a keyname in the request. The key is derived from various source of entropy, including [keyname](https://github.com/babylonlabs-io/finality-provider/blob/743412efe4b37089d4f0921f1190ca43e80f8390/finality-provider/service/app.go#L336-L337). The secret key is used as BTC SK and Babylon SK, and [determines](https://github.com/babylonlabs-io/finality-provider/blob/743412efe4b37089d4f0921f1190ca43e80f8390/finality-provider/service/app.go#L361-L375) the FP's Babylon addr. The PoP will be signed over this Babylon addr.
- however, when starting the FP program it uses the key with key name specified in [config](https://github.com/babylonlabs-io/finality-provider/blob/743412efe4b37089d4f0921f1190ca43e80f8390/clientcontroller/babylon.go#L54-L60). The PoP will be verified where the corresponding Babylon addr in config is treated as signer.

If the keyname in config does not match the keyname specified by the user who sends createFinalityProviderRequest , PoP verification will fail.

A temporary workaround is to always use the fp name given in config file when creating finality provider.

A proper fix will be more complicated. My idea was as follows but there could be simpler ways

- Let keyring in the App struct to only manage BTC keys but not Babylon keys. The Babylon client is already parsing the private key to a keyring in memory so this has no issue
- Make create-finality-provider a one-time thing, e.g., after invoking it once, subsequent invocations to this will return error directly. For example, it can be revamped to sth like fpd initThis is in line with the decision of removing the support of multiple FP instances.
- The handler of create-finality-provider should get Babylon address from Babylon client, rather than creating a new key in the keyring

PR #52, opened 2024-09-12
chore: keep sync fp status

RafilxTenfen:
- Removed previous `MinutesToWaitForConsumer` added in #50 since it was not good enough to wait for the chain to be up, without starting the fpd server to receive fpd creation 
- Add new loop that verifies the FP status on chain and update it accordingly (before it was only one time at startup)
- Add check for keyring key is valid at start of `NewBabylonController` to avoid panic at `mustGetTxSigner`

### DIFF
--- a/clientcontroller/babylon.go
+++ b/clientcontroller/babylon.go
@@ -3,6 +3,7 @@ package clientcontroller
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	sdkErr "cosmossdk.io/errors"
@@ -276,6 +277,12 @@ func (bc *BabylonController) QueryFinalityProviderVotingPower(fpPk *btcec.Public
 		blockHeight,
 	)
 	if err != nil {
+		allowedErr := fmt.Sprintf("rpc error: code = Unknown desc = %s: unknown request", btcstakingtypes.ErrVotingPowerTableNotUpdated.Wrapf("height: %d", blockHeight).Error())
+		if strings.EqualFold(err.Error(), allowedErr) {
+			// if nothing was updated in the voting power table, it should consider as zero VP to start to send pub random
+			return 0, nil
+		}
+
 		return 0, fmt.Errorf("failed to query Finality Voting Power at Height: %w", err)
 	}
 
--- a/clientcontroller/babylon.go
+++ b/clientcontroller/babylon.go
@@ -3,7 +3,6 @@ package clientcontroller
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
 	sdkErr "cosmossdk.io/errors"
@@ -276,14 +275,8 @@ func (bc *BabylonController) QueryFinalityProviderVotingPower(fpPk *btcec.Public
 		bbntypes.NewBIP340PubKeyFromBTCPK(fpPk).MarshalHex(),
 		blockHeight,
 	)
-
 	if err != nil {
-		allowedErr := fmt.Sprintf("rpc error: code = Unknown desc = %s: unknown request", btcstakingtypes.ErrVotingPowerTableNotUpdated.Wrapf("height: %d", blockHeight).Error())
-		if strings.EqualFold(err.Error(), allowedErr) {
-			// if nothing was updated in the voting power table, it should consider as zero VP to start to send pub random
-			return 0, nil
-		}
-		return 0, fmt.Errorf("failed to query BTC delegations: %w", err)
+		return 0, fmt.Errorf("failed to query Finality Voting Power at Height: %w", err)
 	}
 
 	return res.VotingPower, nil
--- a/clientcontroller/babylon.go
+++ b/clientcontroller/babylon.go
@@ -2,8 +2,8 @@ package clientcontroller
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	sdkErr "cosmossdk.io/errors"
@@ -276,11 +276,13 @@ func (bc *BabylonController) QueryFinalityProviderVotingPower(fpPk *btcec.Public
 		bbntypes.NewBIP340PubKeyFromBTCPK(fpPk).MarshalHex(),
 		blockHeight,
 	)
-	if errors.Is(err, btcstakingtypes.ErrVotingPowerTableNotUpdated) {
-		// if nothing was updated in the voting power table
-		return 0, nil
-	}
+
 	if err != nil {
+		allowedErr := fmt.Sprintf("rpc error: code = Unknown desc = %s: unknown request", btcstakingtypes.ErrVotingPowerTableNotUpdated.Wrapf("height: %d", blockHeight).Error())
+		if strings.EqualFold(err.Error(), allowedErr) {
+			// if nothing was updated in the voting power table, it should consider as zero VP to start to send pub random
+			return 0, nil
+		}
 		return 0, fmt.Errorf("failed to query BTC delegations: %w", err)
 	}
 
--- a/clientcontroller/babylon.go
+++ b/clientcontroller/babylon.go
@@ -276,7 +276,7 @@ func (bc *BabylonController) QueryFinalityProviderVotingPower(fpPk *btcec.Public
 		blockHeight,
 	)
 	if err != nil {
-		return 0, fmt.Errorf("failed to query Finality Voting Power at Height: %w", err)
+		return 0, fmt.Errorf("failed to query Finality Voting Power at Height %d: %w", blockHeight, err)
 	}
 
 	return res.VotingPower, nil
--- a/clientcontroller/babylon.go
+++ b/clientcontroller/babylon.go
@@ -2,6 +2,7 @@ package clientcontroller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -275,6 +276,10 @@ func (bc *BabylonController) QueryFinalityProviderVotingPower(fpPk *btcec.Public
 		bbntypes.NewBIP340PubKeyFromBTCPK(fpPk).MarshalHex(),
 		blockHeight,
 	)
+	if errors.Is(err, btcstakingtypes.ErrVotingPowerTableNotUpdated) {
+		// if nothing was updated in the voting power table
+		return 0, nil
+	}
 	if err != nil {
 		return 0, fmt.Errorf("failed to query BTC delegations: %w", err)
 	}
--- a/clientcontroller/babylon.go
+++ b/clientcontroller/babylon.go
@@ -3,7 +3,6 @@ package clientcontroller
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
 	sdkErr "cosmossdk.io/errors"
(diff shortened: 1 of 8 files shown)


finality-provider/store/storedfp.go
```diff
@@ -77,3 +77,14 @@ func (sfp *StoredFinalityProvider) ToFinalityProviderInfo() *proto.FinalityProvi
 		Status:          sfp.Status.String(),
 	}
 }
+
+// ShouldSyncStatusFromVotingPower returns true if it should modify the status based on the
+// blockchain data.
```
Lazar955, line 82:
```suggestion
// ShouldSyncStatusFromVotingPower returns true if the status should be updated 
// based on the provided voting power or the current status of the finality provider.
// 
// It returns true if the voting power is greater than zero, or if the status
// is either 'CREATED' or 'ACTIVE'.
```

Lazar955 (approved):
lgtm!

clientcontroller/babylon.go
```diff
@@ -274,7 +277,13 @@ func (bc *BabylonController) QueryFinalityProviderVotingPower(fpPk *btcec.Public
...
 	)
 	if err != nil {
-		return 0, fmt.Errorf("failed to query BTC delegations: %w", err)
+		allowedErr := fmt.Sprintf("rpc error: code = Unknown desc = %s: unknown request", btcstakingtypes.ErrVotingPowerTableNotUpdated.Wrapf("height: %d", blockHeight).Error())
+		if strings.EqualFold(err.Error(), allowedErr) {
+			// if nothing was updated in the voting power table, it should consider as zero VP to start to send pub random
+			return 0, nil
+		}
+
+		return 0, fmt.Errorf("failed to query Finality Voting Power at Height: %w", err)
```
gitferry, line 286: I started to think we should not ignore this error because the error essentially means that there's no voting power for any fps yet. If we keep retrying, then eventually the error will be gone. If we ignore this error, we won't revisit the voting power for this height, which could lead to missing a block that the FP has voting power.

I think we should ensure that `QueryFinalityProviderVotingPower` would not be called before [waitForActivation](https://github.com/babylonlabs-io/finality-provider/blob/f7c17b17e6ec546ac9370b9a3d21686eeb5e12a7/finality-provider/service/chain_poller.go#L205) is finished.

gitferry (commented):
Maybe I was missing something but I don't understand the motivation of having a loop for state sync. What specific case does this PR wants to resolve?

clientcontroller/babylon.go
```diff
@@ -47,10 +48,6 @@ func NewBabylonController(
 
 	bbnConfig := fpcfg.BBNConfigToBabylonConfig(cfg)
 
-	if err := bbnConfig.Validate(); err != nil {
-		return nil, fmt.Errorf("invalid config for Babylon client: %w", err)
-	}
```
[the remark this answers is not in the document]
RafilxTenfen: this is already being validated at `bbnclient.New(`

clientcontroller/babylon.go
RafilxTenfen: humm, this makes sense 
But I had understood that the finality provider should send pub rand list even if he does not have voting power, only the finality signature should verify that it has voting power, before sending 

Is my understanding correct?

RafilxTenfen:
> Maybe I was missing something but I don't understand the motivation of having a loop for state sync. What specific case does this PR wants to resolve?

This PR wants to solve: 
- FPs created offline following the [docs](https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/17662054/On-chain+Registry+Preparation) that will be included in the upgrade, do not pass to the normal process of finality provider registration where it uses the command `fpd register-finality-provider`, since they will already be on the chain state. So the finality provider must keep asking the chain "Hey, are you ready? If so, give me the registered finality providers and I will start the fpd instance"
- Finality providers to be able to turn on and start the `fpd` even before the consumer chain is running, and when the consumer chain becomes available and ready, it starts sending the needed messages

@gitferry

clientcontroller/babylon.go
gitferry: Yes. Thing is that if the fp gets `0` from `QueryFinalityProviderVotingPower`, it will skip sending finality votes for this height for good. In this PR, `QueryFinalityProviderVotingPower` can return `0` due to the protocol is not activated. So it is possible that `QueryFinalityProviderVotingPower` will return non-zero value for this height after the protocol is activated. I was worried that the fp might miss such heights due to this change

RafilxTenfen: humm, I think I understand 
For finality signatures we still want to respect this error and avoid skipping it! 
For send the pub rand list, we should still send it even if the btcstaking protocol is not activated yet

RafilxTenfen: modified at https://github.com/babylonlabs-io/finality-provider/pull/52/commits/a43b3f1865f7363630f9b69aa4112f51e4623056, now only the portion that updates from Created to Register, allows the error `ErrVotingPowerTableNotUpdated`

clientcontroller/babylon.go
```diff
@@ -274,7 +276,7 @@ func (bc *BabylonController) QueryFinalityProviderVotingPower(fpPk *btcec.Public
 		blockHeight,
 	)
 	if err != nil {
-		return 0, fmt.Errorf("failed to query BTC delegations: %w", err)
+		return 0, fmt.Errorf("failed to query Finality Voting Power at Height: %w", err)
```
gitferry, line 279:
```suggestion
		return 0, fmt.Errorf("failed to query Finality Voting Power at Height %d: %w", blockHeight, err)
```

gitferry (commented):
> * FPs created offline following the [docs](https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/17662054/On-chain+Registry+Preparation) that will be included in the upgrade, do not pass to the normal process of finality provider registration where it uses the command fpd register-finality-provider, since they will already be on the chain state. So the finality provider must keep asking the chain "Hey, are you ready? If so, give me the registered finality providers and I will start the fpd instance"
> * Finality providers to be able to turn on and start the fpd even before the consumer chain is running, and when the consumer chain becomes available and ready, it starts sending the needed messages

I'm still not convinced that we should use a long-lived loop for the above. I think the only job of this loop is to wait for the status of the fp to be `>= REGISTERED && < SLASHED`. After the fp is started then the loop runs for nothing (we already have `monitorStatusUpdate` to periodically update fp status). Why can't we use a one-time `waitForChainReady` and `waitForActivation` before initiating the fp instance? Note that the poller starts after the protocol is [activated](https://github.com/babylonlabs-io/finality-provider/blob/f7c17b17e6ec546ac9370b9a3d21686eeb5e12a7/finality-provider/service/chain_poller.go#L205), meaning that the first block sent from the poller should already be activated.

clientcontroller/babylon.go
RafilxTenfen: done at https://github.com/babylonlabs-io/finality-provider/pull/52/commits/94b8ff69a45d7d13a1f4ffa9f392a33d2cb95b20

RafilxTenfen:
> > * FPs created offline following the [docs](https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/17662054/On-chain+Registry+Preparation) that will be included in the upgrade, do not pass to the normal process of finality provider registration where it uses the command fpd register-finality-provider, since they will already be on the chain state. So the finality provider must keep asking the chain "Hey, are you ready? If so, give me the registered finality providers and I will start the fpd instance"
> > * Finality providers to be able to turn on and start the fpd even before the consumer chain is running, and when the consumer chain becomes available and ready, it starts sending the needed messages
> 
> I'm still not convinced that we should use a long-lived loop for the above. I think the only job of this loop is to wait for the status of the fp to be `>= REGISTERED && < SLASHED`. After the fp is started then the loop runs for nothing (we already have `monitorStatusUpdate` to periodically update fp status). Why can't we use a one-time `waitForChainReady` and `waitForActivation` before initiating the fp instance? Note that the poller starts after the protocol is [activated](https://github.com/babylonlabs-io/finality-provider/blob/f7c17b17e6ec546ac9370b9a3d21686eeb5e12a7/finality-provider/service/chain_poller.go#L205), meaning that the very first block sent from the poller should already be activated.

Without this loop `syncChainFpStatusLoop` imagine that the `fpd` starts by complete, which starts to listen for requests like (create and register FP). We call `fpd create-finality-provider` to store the new FP into the fpd database, we can't call `fpd register-finality-provider`, because the FPs inserted during the upgrade are already registered on chain, the `fpd` just doesn't know they are already registered (and will never know without checking the chain). 
So, no FP instance will start, because they are not considered as `REGISTERED` 


One thing is that I do agree, after the FP is marked as REGISTERED and the instance start, there is no use for this loop and it could be stopped and from #34 we are not going to support multiple fps, so I will modify that after the FP is considered registered, the loop  `syncChainFpStatusLoop` is stopped.

What do you think? @gitferry

gitferry:
> Without this loop syncChainFpStatusLoop imagine that the fpd starts by complete, which starts to listen for requests like (create and register FP). We call fpd create-finality-provider to store the new FP into the fpd database, we can't call fpd register-finality-provider, because the FPs inserted during the upgrade are already registered on chain, the fpd just doesn't know they are already registered (and will never know without checking the chain).
So, no FP instance will start, because they are not considered as REGISTERED

I understand the use case is to sync the fp status from CREATED to REGISTERED without the user having to run the register command. My question is why is the previous implementation not enough to cover this case? See https://github.com/babylonlabs-io/finality-provider/blob/d06cde51674f518d54aaabf92a17df54596dfadf/finality-provider/service/app.go#L292.

RafilxTenfen:
> to sync the fp status from CREATED to REGISTERED without the user having to run the register command. My question is why is the previous implementation not enough to cover this case? See

The old implementation was waiting for the chain to be ready to update from CREATED to REGISTERED to start the rest of loops (but this also did not started the fp server and handler of requests). The FP server needed to be online to receive the call `fpd create-finality-provider`

finality-provider/service/app.go
```diff
@@ -237,73 +240,65 @@ func (app *FinalityProviderApp) getFpPrivKey(fpPk []byte) (*btcec.PrivateKey, er
...
 	}
 
 	for _, fp := range fps {
 		vp, err := app.cc.QueryFinalityProviderVotingPower(fp.BtcPk, latestBlock.Height)
 		if err != nil {
-			// if error occured then the finality-provider is not registered in the Babylon chain yet
+			// if error occured then the finality-provider is not registered in the Babylon chain yet or
+			// there is nothing in the voting power table, so it should not start the fp.
+			allowedErr := fmt.Sprintf("failed to query Finality Voting Power at Height %d: rpc error: code = Unknown desc = %s: unknown request", latestBlock.Height, bstypes.ErrVotingPowerTableNotUpdated.Wrapf("height: %d", latestBlock.Height).Error())
+			if !strings.EqualFold(err.Error(), allowedErr) {
```
gitferry, line 261:
```suggestion
			if !errors.Is(err, bstypes.ErrVotingPowerTableNotUpdated) {
```

gitferry:
> The old implementation was waiting for the chain to be ready to update from CREATED to REGISTERED to start the rest of loops (but this also did not started the fp server and handler of requests). The FP server needed to be online to receive the call fpd create-finality-provider

sorry I edited my comment before your reply. Yeah, now I think I slowly get it. Thanks!

finality-provider/service/app.go
```diff
@@ -682,3 +678,32 @@ func (app *FinalityProviderApp) metricsUpdateLoop() {
 		}
 	}
 }
+
+func (app *FinalityProviderApp) syncChainFpStatusLoop() {
```
gitferry, line 682: we should add a doc string saying that this loop will exit after at least one fp instance is started

gitferry (approved):
Great work! Thanks for all the changes and explanation! Only one missing piece is comprehensive tests, especially on the transition scenario. This can be done in a separate PR

finality-provider/service/app.go
RafilxTenfen: I was not able to do it with `error.Is` due to wrapping of rpc error of cosmos RPC handler

finality-provider/service/app.go
RafilxTenfen: done at https://github.com/babylonlabs-io/finality-provider/pull/52/commits/502182975036de19d35550c2a14e73bc89c80644

RafilxTenfen:
I will be adding tests in a next PR

Issue #53, opened 2024-09-15
FP submission loop stuck when poller `blockInfoChan` is full


bap2pecs:
## Problem

we deployed our own L2 OP devnet with a private Babylon network. The consumer FP ran out of BBN token last week. Then we topped up the balance and restarted the FP. The issue was found and resolved after ~12 hours.

After, the FP started to fast sync to catch up submitting signatures. We saw fast sync finished in the log:

```
2024-09-08T16:38:03.271817Z	info	fast sync is finished	{"pk": "...", "synced_height": 46346, "last_processed_height": 46346}
```

Then the FP stopped to submit new signatures from block 46347. We found out that there is no logs of “the finality-provider received a new block, start processing” (in finalitySigSubmissionLoop() ). So it seems there was sth wrong with the chain poller.

Then we noticed that fp.poller.SkipToHeight() was called after fp.logger.Info("fast sync is finished",…)

But it’s weird that we don’t see any logs of “the poller has skipped height(s)“.

## Root cause analysis

we found out that the code is stuck here

```
func (cp *ChainPoller) SkipToHeight(height uint64) error {
	...
	select {
	case <-cp.quit:
		return fmt.Errorf("the chain poller is stopped")
	case cp.skipHeightChan <- &skipHeightRequest{height: height, resp: respChan}:
	}

        // stuck here
	select {
	case <-cp.quit:
		return fmt.Errorf("the chain poller is stopped")
	case resp := <-respChan:
		return resp.err
	}
}
```

this is what have happened:

- after topping up fund, `finalitySigSubmissionLoop()` entered fast sync mode to catch up to the latest block
- at the same time, `pollChain()` runs in a separate process to poll new blocks and put into the `blockInfoChan` channel
- before fast-sync is finished, the `blockInfoChan` got full. so `pollChain()` got stuck
- so `case req := <-cp.skipHeightChan:` won't be called.
- in FP, when fast sync finished, poller `SkipToHeight()` will be called
- and it's stuck at the last `select` statement b/c `respChan` is never triggered
- this is b/c `respChan` will be used here:
```
		case req := <-cp.skipHeightChan:
			// no need to skip heights if the target height is not higher
			// than the next height to retrieve
			targetHeight := req.height
			if targetHeight <= cp.nextHeight {
				resp := &skipHeightResponse{
					err: fmt.Errorf(
						"the target height %d is not higher than the next height %d to retrieve",
						targetHeight, cp.nextHeight)}
[...]

PR #54, opened 2024-09-15
Start poller after fast sync is completed

bap2pecs:
## Summary

this is to fix https://github.com/babylonlabs-io/finality-provider/issues/53

it also makes sense b/c fast sync will fetch the blocks anyway. so it's a waste of resources as well if there is another poller routine doing the same thing

## Test Plan
```
make test-e2e-op
make test
make test-e2e-babylon --run ^TestFastSync$
```


finality-provider/service/chain_poller.go
```diff
@@ -97,6 +97,10 @@ func (cp *ChainPoller) Start(startHeight uint64) error {
 }
 
 func (cp *ChainPoller) Stop() error {
+	if !cp.IsRunning() {
```
bap2pecs, line 100: without this, `TestOpFpNoVotingPower` will fail b/c it's trying to stop a poller that's not started

finality-provider/service/fp_instance.go
```diff
@@ -116,12 +116,22 @@ func (fp *FinalityProviderInstance) Start() error {
...
-		return fmt.Errorf("failed to start the poller: %w", err)
+	// get the last finalized height
+	lastFinalizedBlock, err := fp.latestFinalizedBlockWithRetry()
+	if err != nil {
+		return err
 	}
 
-	fp.poller = poller
+	// Start the poller if fast sync is disabled or there's no finalized block
+	if (fp.cfg.FastSyncInterval == 0 || lastFinalizedBlock == nil) && !fp.poller.IsRunning() {
```
bap2pecs, line 128: without `lastFinalizedBlock == nil`, babylon e2e `TestFastSync` will fail b/c `tm.WaitForFpVoteCast(t, fpIns)` will timeout

this is b/c in the code below, `tryFastSync` will not do anything

image

bap2pecs:
not sure why CI failed even on lint. `make lint` runs fine locally

image

bap2pecs:
also seems babylon-private access in CI is still broken

image

SebastianElvis:
Thanks for the proposed fix! @gitferry what do you think about starting poller only after fast sync? actually I remember we had a discussion on it but somehow did not push it further

Also, if this idea is approved, how about merging it to `main` rather than the base branch?

finality-provider/service/chain_poller.go
gitferry: isn't this the same as `!cp.isStarted.Swap(false)`? It's just nil is returned instead of an error

gitferry (commented):
In the previous implementation, the poller is started after `bootstrap` is finished which involves `fast sync`, so I don't understand what specific case does this PR want to address. Did I miss anything?

bap2pecs:
> In the previous implementation, the poller is started after `bootstrap` is finished which involves `fast sync`, so I don't understand what specific case does this PR want to address. Did I miss anything?

this is to fix an edge case where FP runs out of BBN token to vote and couldn't recover. see details in https://github.com/babylonlabs-io/finality-provider/issues/53

bap2pecs:
@gitferry. ok we just took another look at the logs. so what happened was:

after topping up BBN and restarting the FP, `bootstrap` will only proceed to the block that has pub rand. when it detects a block that doesn't have PR, it will return without error:

```
func (fp *FinalityProviderInstance) hasRandomness(b *types.BlockInfo) (bool, error) {
	lastCommittedHeight, err := fp.GetLastCommittedHeight()
	if err != nil {
		return false, err
	}
	if b.Height > lastCommittedHeight {
		fp.logger.Debug(
			"the finality provider has not committed public randomness for the height",
			zap.String("pk", fp.GetBtcPkHex()),
			zap.Uint64("block_height", b.Height),
			zap.Uint64("last_committed_height", lastCommittedHeight),
		)
		return false, nil
	}

	return true, nil
}
```

then it will consider bootstrap complete and prints "the finality-provider has been bootstrapped"

see relevant logs
```
2024-09-08T13:06:49.101779Z	info	Starting finality-provider instance	{"pk": "e1241d1fa4087b853bf8f92734241e52afddd629fdb5620e91c393a907425e5e"}
2024-09-08T13:06:49.104649Z	debug	the finality-provider is entering fast sync	{"start_height": 25444, "target_block_height": 46340}
2024-09-08T13:07:53.076012Z	debug	the finality provider has not committed public randomness for the height	{"pk": "e1241d1fa4087b853bf8f92734241e52afddd629fdb5620e91c393a907425e5e", "block_height": 46334, "last_committed_height": 25500}
2024-09-08T13:07:53.090461Z	info the finality-provider has been bootstrapped {"pk": "e1241d1fa4087b853bf8f92734241e52afddd629fdb5620e91c393a907425e5e", "height": 25501}
```

cc @SebastianElvis

finality-provider/service/fp_instance.go
SebastianElvis: Can we move `!fp.poller.IsRunning()` out to a new if condition? imo if `!fp.poller.IsRunning()` then we need to `return err`? Also if this if statement is false, when will the poller start?

SebastianElvis (approved):
It would be good if @gitferry can take a final look :+1:

finality-provider/service/fp_instance.go
```diff
@@ -275,6 +285,16 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
 					zap.Uint64("last_processed_height", res.LastProcessedHeight),
 				)
 
+				// start poller after fast sync is finished
+				if !fp.poller.IsRunning() {
```
gitferry, line 289: the poller is set after `bootstrap` is called. Won't this panic here?

gitferry:
> @gitferry. ok we just took another look at the logs. so what happened was:
> 
> after topping up BBN and restarting the FP, `bootstrap` will only proceed to the block that has pub rand. when it detects a block that doesn't have PR, it will return without error:
> 
> ```
> func (fp *FinalityProviderInstance) hasRandomness(b *types.BlockInfo) (bool, error) {
> 	lastCommittedHeight, err := fp.GetLastCommittedHeight()
> 	if err != nil {
> 		return false, err
> 	}
> 	if b.Height > lastCommittedHeight {
> 		fp.logger.Debug(
> 			"the finality provider has not committed public randomness for the height",
> 			zap.String("pk", fp.GetBtcPkHex()),
> 			zap.Uint64("block_height", b.Height),
> 			zap.Uint64("last_committed_height", lastCommittedHeight),
> 		)
> 		return false, nil
> 	}
> 
> 	return true, nil
> }
> ```
> 
> then it will consider bootstrap complete and prints "the finality-provider has been bootstrapped"
> 
> see relevant logs
> 
> ```
> 2024-09-08T13:06:49.101779Z	info	Starting finality-provider instance	{"pk": "e1241d1fa4087b853bf8f92734241e52afddd629fdb5620e91c393a907425e5e"}
> 2024-09-08T13:06:49.104649Z	debug	the finality-provider is entering fast sync	{"start_height": 25444, "target_block_height": 46340}
> 2024-09-08T13:07:53.076012Z	debug	the finality provider has not committed public randomness for the height	{"pk": "e1241d1fa4087b853bf8f92734241e52afddd629fdb5620e91c393a907425e5e", "block_height": 46334, "last_committed_height": 25500}
> 2024-09-08T13:07:53.090461Z	info the finality-provider has been bootstrapped {"pk": "e1241d1fa4087b853bf8f92734241e52afddd629fdb5620e91c393a907425e5e", "height": 25501}
> ```
> 
> cc @SebastianElvis

It seems that the issue is that fast sync reaches the latest height that pub rand is committed but still far from the tip height, so the fast sync got triggered frequently causing the poller stuck. If it is true then the issue should be why the pub rand is not caught up because we have a separate goroutine to commit pub rand continuously.

bap2pecs:
> If it is true then the issue should be why the pub rand is not caught up because we have a separate goroutine to commit pub rand continuously.

@lesterli could you help take a look later today? i will be mostly in events today

lesterli:
> > If it is true then the issue should be why the pub rand is not caught up because we have a separate goroutine to commit pub rand continuously.
> 
> @lesterli could you help take a look later today? i will be mostly in events today

after checking the FP codes, the `defaultRandomInterval` is `30s` and `defaultFastSyncInterval` is `10s`, when fast sync was running and there were many lagging blocks(`25444-46340`),  it is possible that some blocks' pub rand had not been submitted.

lesterli:
> `bootstrap` is finished which involves `fast sync`

@gitferry in the `bootstrap()`, should we first check `FastSyncInterval`? If the fast sync is disabled, we should not trigger fast sync, right?

finality-provider/service/fp_instance.go
lesterli: If this `if` statement is `false`, we start the poller after fast sync is finished.
 ```Go
 // start poller after fast sync is finished
if !fp.poller.IsRunning() {
	err := fp.poller.Start(res.LastProcessedHeight + 1)
	if err != nil {
		fp.logger.Error("failed to start the poller", zap.Error(err))
		fp.reportCriticalErr(err)
	}
	continue
}
 ```

finality-provider/service/chain_poller.go
lesterli: yes, now `TestOpFpNoVotingPower` will start the poller, it can be removed.

bap2pecs:
@gitferry so the problem is fast sync cannot fast-sync pub randomnesses. and it doesn't make sense to fast-sync pub randomnesses either b/c iirc we need to wait for some time for a pub rand to be useful.

so it seems we will still need this change

gitferry:
> @gitferry in the bootstrap(), should we first check FastSyncInterval? If the fast sync is disabled, we should not trigger fast sync, right?

Yes, we have the check at the beginning of `checkLaggingLoop`

finality-provider/service/chain_poller.go
```diff
@@ -74,7 +74,8 @@ func NewChainPoller(
 
 func (cp *ChainPoller) Start(startHeight uint64) error {
 	if cp.isStarted.Swap(true) {
-		return fmt.Errorf("the poller is already started")
+		cp.logger.Info("the poller is already started")
+		return nil
```
gitferry, line 78: Why change to return `nil`? If so, it will hide the fact that the poller is called twice, which we do not expect.

finality-provider/service/fp_instance.go
gitferry: Not sure we should add these check because up to here the fast sync is already finished and should be safe to start the poller. Can we have a workaround in e2e?

finality-provider/service/fp_instance.go
```diff
@@ -275,6 +219,16 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
...
+					err := fp.poller.Start(res.LastProcessedHeight + 1)
+					if err != nil {
+						fp.logger.Error("failed to start the poller", zap.Error(err))
+						fp.reportCriticalErr(err)
+					}
+					continue
+				}
```
gitferry, line 230: I don't think the start of the poller should be controlled within the sig submission loop. The poller should be already started when the loop is on

gitferry:
> @gitferry so the problem is fast sync cannot fast-sync pub randomnesses. and it doesn't make sense to fast-sync pub randomnesses either b/c iirc we need to wait for some time for a pub rand to be useful.
> 
> so it seems we will still need this change

we have a separate loop continuously committing pub rand so I think eventually fast-sync will go to the expected height

bap2pecs:
> we have a separate loop continuously committing pub rand so I think eventually fast-sync will go to the expected height

right but the problem is bootstrap () will not wait for it. thus the edge case i.e. poller channel gets full still exists. and this PR tries to fix that

bap2pecs:
ok we just realized that this PR still has some problems.

https://github.com/babylonlabs-io/finality-provider/compare/base/consumer-chain-support...fix/tmp-fix-poller?expand=1 this might be an easier fix to avoid all the headaches but still resolves #53 

cc @gitferry @SebastianElvis

bap2pecs:
another discussion topic is whether we should fast sync pub rand

lesterli:
> * @lesterli will cherry-pick [feat: batch processing non-finalised blocks #57](https://github.com/babylonlabs-io/finality-provider/pull/57) to base

created https://github.com/babylonlabs-io/finality-provider/pull/77

bap2pecs:
closing as we have removed fast sync

Issue #55, opened 2024-09-17
OP: fast sync got constantly triggered


lesterli:
## Problem

In our deployed L2 OP devnet, we found fast sync got constantly triggered every interval (`20s`), we have below configurations:
* OP L2 block time: 2s
* FP chain poller interval: 2s
* FP fast sync interval: 20s
* FP fast sync gap blocks: 6

## Root cause analysis

After debugging we found the root cause is: within the fast sync interval-20s, there are 10 new blocks, but each EOTS sig submission takes 5 sec, FP can only manage to submit 4 blocks, so 6 blocks got left behind, and the fast sync gap is 6. so fast sync got triggered every time.

So this means for chains that produce blocks very fast, fast sync will be constantly triggered. 

## Proposed fix

Batch processing all queued blocks in the `finalitySigSubmissionLoop` instead of processing them one by one.

```
func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
    for {
		select {
		case b := <-fp.poller.GetBlockInfoChan():
        // res, err := fp.retrySubmitFinalitySignatureUntilBlockFinalized(&nextBlock)
        res, err := fp.retryBatchSubmitFinalitySignatureUntilBlockFinalized(&queuedBlocks)
```

bap2pecs:
follow-up tasks:

- [ ] verify how much improvements it has
- [ ] add automated test

bap2pecs:
fast sync is removed

PR #57, opened 2024-09-18
feat: batch processing non-finalised blocks

lesterli:
## Summary

This PR fixes the issue https://github.com/babylonlabs-io/finality-provider/issues/55

## Test Plan

```
make lint
make test
make test-e2e-babylon --run ^TestFastSync$
make test-e2e-op
```


finality-provider/service/fp_instance.go
```diff
@@ -194,52 +194,81 @@ func (fp *FinalityProviderInstance) IsRunning() bool {
...
+					}
+					// the block is finalized, no need to submit finality signature
+					if isFinalized {
+						fp.MustSetLastProcessedHeight(b.Height)
+						continue
+					}
+
+					pollerBlocks = append(pollerBlocks, b)
+				default:
+					goto processBlocks
```
SebastianElvis, line 259: Is there a way to avoid the `goto` statement, as it seems not a good practice.

Here `default` is hit when the channel becomes empty. Can we replace this `goto` with `break` then?

SebastianElvis (approved):
This is an algorithm-level change so @gitferry what do you think about submitting finality signatures for all blocks if multiple blocks are in the channel?

Also, if this idea is approved, how about we merge it to `main`?

bap2pecs:
> Also, if this idea is approved, how about we merge it to `main`?

yes and also #54. we can cherry pick the PRs

finality-provider/service/fp_instance.go
lesterli: replace `goto` statement and use a boolean `fetchMoreBlocks`.

finality-provider/service/fp_instance.go
```diff
@@ -194,52 +193,51 @@ func (fp *FinalityProviderInstance) IsRunning() bool {
 func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
 	defer fp.wg.Done()
 
+	var targetHeight uint64
 	for {
 		select {
 		case b := <-fp.poller.GetBlockInfoChan():
```
gitferry, line 199: I think we can put it in the default case where we have `getAllBlocksFromChan()` to drain the channel and then try to process each block and do batch send

finality-provider/service/fp_instance.go
```diff
@@ -194,52 +193,51 @@ func (fp *FinalityProviderInstance) IsRunning() bool {
...
-
-			// use the copy of the block to avoid the impact to other receivers
-			nextBlock := *b
-			res, err := fp.retrySubmitFinalitySignatureUntilBlockFinalized(&nextBlock)
+			// Fetch all available blocks
+			// Note: not all the blocks in the range will have votes cast
+			// due to lack of voting power or public randomness, so we may
+			// have gaps during processing
+			pollerBlocks := []*types.BlockInfo{b}
+			fetchMoreBlocks := true
```
gitferry, line 215: we probably don't need this flag. We can use `for {}` and inject `break` in `default` case

gitferry (commented):
In general I like the batching solution. Left some suggestion in code-wise. We'd better add a test case to ensure the solution works

finality-provider/service/fp_instance.go
lesterli: Do you mean to put the `var targetHeight uint64` in the `default` case of the inner loop?

gitferry: I meant the entire block fetching logic can be put in the `default` case so that we can have a method to fetch them all other than `fetch one, and the rest`

lesterli: Got it

finality-provider/service/fp_instance.go
lesterli: yes, it is removed.

gitferry (approved):
LGTM! We should test this case but can be done in a separate PR

bap2pecs:
@lesterli have we tested this using our demo app to see how long it takes after this change?

PR #62, opened 2024-09-23
fix: merge base to main

### LINKED ISSUE #300: merge phase-3 branch into `main`

SebastianElvis:
Resolves #300 

This PR merges base branch to main

NOTE: all comments in this PR are addressed in #315


cosmwasmclient/query/tendermint.go
```diff
@@ -0,0 +1,60 @@
...
+	return c.RPCClient.Subscribe(context.Background(), subscriber, query, outCapacity...)
+}
+
+func (c *QueryClient) Unsubscribe(subscriber, query string) error {
+	return c.RPCClient.Unsubscribe(context.Background(), subscriber, query)
+}
+
+func (c *QueryClient) UnsubscribeAll(subscriber string) error {
+	return c.RPCClient.UnsubscribeAll(context.Background(), subscriber)
+}
```
Lazar955, line 60: let's add ctx as an argument and let the caller decide

itest/babylon/babylon_test_manager.go
```diff
@@ -0,0 +1,355 @@
...
+func (tm *TestManager) WaitForServicesStart(t *testing.T) {
+	require.Eventually(t, func() bool {
+		_, err := tm.BBNClient.QueryBestBlock()
+		return err == nil
+	}, eventuallyWaitTimeOut, eventuallyPollTime)
+
+	t.Logf("Babylon node is started")
+}
+
+func StartManagerWithFinalityProvider(t *testing.T, ctx context.Context) (*TestManager, *service.FinalityProviderInstance) {
```
Lazar955, line 231: can you revert to previous impl where we specify num fps. As in `TestFinalityProviderLifeCycle` we test running 2 fps against single eots manager

itest/bytecode/version.txt
```diff
@@ -0,0 +1 @@
+a33989c5c7c11b882bc7d5f50c1dfe30bae47fa6
```
Lazar955, line 1: What is this file used for?

Lazar955 (commented):
amazing work, will continue reviewing as there's quite a lot of new code for a single session, some comments

clientcontroller/babylon/babylon.go
```diff
@@ -48,13 +52,29 @@ func NewBabylonController(
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Babylon client: %w", err)
+	}
+
+	// makes sure that the key in config really exists and is a valid bech32 addr
+	// to allow using MustGetTxSigner
+	if _, err := bc.GetAddr(); err != nil {
+		return nil, err
+	}
```
Lazar955, line 70: remove this, `GetAddr()` is now checked in the Start, and bbnClient is passed as an argument

itest/bytecode/version.txt
SebastianElvis: This is a file auto-generated when compiling the cosmwasm contract. It's used for identifying the version of the source code corresponding to the contract.

Try out `cargo optimize` under https://github.com/babylonlabs-io/babylon-contract/

.github/workflows/publish.yml
```diff
@@ -15,12 +15,47 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.23'
+      - name: Run e2e BCD tests
+        run: make test-e2e-bcd
+
+# TODO: disable e2e_op as removing fast sync might introduce nontrivial change
```
gitferry, line 45: Do we have an issue to track it? Seems critical

clientcontroller/babylon/babylon.go
```diff
@@ -28,14 +30,16 @@ import (
...
 	fpcfg "github.com/babylonlabs-io/finality-provider/finality-provider/config"
 	"github.com/babylonlabs-io/finality-provider/types"
 )
 
-var _ ClientController = &BabylonController{}
+var _ api.ClientController = &BabylonController{}
 
 var emptyErrs = []*sdkErr.Error{}
 
+//nolint:revive
```
gitferry, line 42: what does this mean?

clientcontroller/babylon/babylon.go
```diff
@@ -118,6 +138,7 @@ func (bc *BabylonController) reliablySendMsgs(msgs []sdk.Msg, expectedErrs []*sd
 // RegisterFinalityProvider registers a finality provider via a MsgCreateFinalityProvider to Babylon
 // it returns tx hash and error
 func (bc *BabylonController) RegisterFinalityProvider(
+	chainID string,
```
gitferry, line 141: nit: do we want to check `chainID` is not empty before sending the tx?

clientcontroller/babylon/babylon.go
```diff
@@ -268,22 +290,15 @@ func (bc *BabylonController) UnjailFinalityProvider(fpPk *btcec.PublicKey) (*typ
...
-}
-
-// QueryFinalityProviderSlashedOrJailed - returns if the fp has been slashed, jailed, err
-func (bc *BabylonController) QueryFinalityProviderSlashedOrJailed(fpPk *btcec.PublicKey) (bool, bool, error) {
-	fpPubKey := bbntypes.NewBIP340PubKeyFromBTCPK(fpPk)
-	res, err := bc.bbnClient.QueryClient.FinalityProvider(fpPubKey.MarshalHex())
-	if err != nil {
-		return false, false, fmt.Errorf("failed to query the finality provider %s: %w", fpPubKey.MarshalHex(), err)
+	if res == nil {
+		return &types.TxResponse{}, nil
```
gitferry, line 294: how could this case happen?

clientcontroller/babylon/babylon.go
gitferry, line 1: seems there are many unused code in this file. We should remove them

metrics/tools/go.mod
gitferry, line 1: why do we need this?

metrics/tools/tools.go
gitferry, line 1: do we still need this as we are using babylon container?

tools/go.mod
gitferry, line 1: same here

types/pub_rand_commit.go
```diff
@@ -1,11 +1,34 @@
...
+	if prc.NumPubRand < 1 {
+		return fmt.Errorf("NumPubRand must be >= 1, got %d", prc.NumPubRand)
+	}
+
+	return nil
+}
+
+// EndHeight returns the last height covered by this public randomness commitment
+func (prc *PubRandCommit) EndHeight() uint64 {
+	return prc.StartHeight + prc.NumPubRand - 1
```
gitferry, line 29: what if `prc.StartHeight == 0` and  `prc.NumPubRand == 0`?

gitferry (approved):
Great work! I didn't look into details about consumer implementations. Comments mostly about some questions and cleanups. No blocker from my side.

gitferry:
Is this compatible with phase-2?

clientcontroller/babylon/babylon.go
SebastianElvis: It's a linter rule for some naming consistency check. If we remove this line then linter will fail with the following

```
➜  finality-provider git:(base/consumer-chain-support) golangci-lint run 
clientcontroller/babylon/babylon.go:42:6: exported: type name will be used as babylon.BabylonController by other packages, and that stutters; consider calling this Controller (revive)
type BabylonController struct {
     ^
```

For now we surpassed such naming convention issues. We can revise naming afterwards. Wdyt?

clientcontroller/babylon/babylon.go
SebastianElvis: If empty this it means the FP is a Babylon FP ([ref](https://github.com/babylonlabs-io/babylon/blob/2108dce3b7673b0f0d545ec1df13432f89799940/x/btcstaking/keeper/finality_providers.go#L36-L42)). Added the docstring to make it clear

clientcontroller/babylon/babylon.go
SebastianElvis: This was actually introduced at https://github.com/babylonlabs-io/finality-provider/pull/124 by you :sweat_smile: Do you remember any context?

.github/workflows/publish.yml
SebastianElvis: This is fixed actually. Reenabled

clientcontroller/babylon/babylon.go
SebastianElvis: Good catch! Removed a bunch of functions

metrics/tools/go.mod
SebastianElvis: Good catch, it's strange to have this file under `metrics/`. It belongs to `tools/` folder but not here. Maybe there was some mistake copything things here and there :( removed

metrics/tools/tools.go
SebastianElvis: We were still using it for the e2e test of Cosmos integration ([ref](https://github.com/babylonlabs-io/finality-provider/blob/c6f6fc1c88287d3eb995f2ce561f8069e8b04ffa/tools/go.mod#L7-L8)). Better dockerise the `bcd` container as well. Created https://github.com/babylonlabs-io/finality-provider/issues/322 to track this issue.

types/pub_rand_commit.go
SebastianElvis: `NumPubRand` should be at least 1 and `Validate()` checks this [ref](https://github.com/babylonlabs-io/finality-provider/blob/c6f6fc1c88287d3eb995f2ce561f8069e8b04ffa/types/pub_rand_commit.go#L17-L26). The check is performed in every controller ([Babylon](https://github.com/babylonlabs-io/finality-provider/blob/c6f6fc1c88287d3eb995f2ce561f8069e8b04ffa/clientcontroller/babylon/consumer.go#L349-L350), [Cosmos](https://github.com/babylonlabs-io/finality-provider/blob/c6f6fc1c88287d3eb995f2ce561f8069e8b04ffa/clientcontroller/cosmwasm/consumer.go#L323-L324), [OP stack](https://github.com/babylonlabs-io/finality-provider/blob/c6f6fc1c88287d3eb995f2ce561f8069e8b04ffa/clientcontroller/opstackl2/consumer.go#L508-L509)).

cosmwasmclient/client/client.go
```diff
@@ -0,0 +1,101 @@
...
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	wasmdparams "github.com/CosmWasm/wasmd/app/params"
+	"github.com/babylonlabs-io/finality-provider/cosmwasmclient/config"
+	"github.com/babylonlabs-io/finality-provider/cosmwasmclient/query"
+	rpchttp "github.com/cometbft/cometbft/rpc/client/http"
+	"github.com/cosmos/relayer/v2/relayer/chains/cosmos"
```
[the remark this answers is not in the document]
SebastianElvis: Good point, fixed in https://github.com/babylonlabs-io/finality-provider/pull/315

SebastianElvis:
> Is this compatible with phase-2?

I think so, as the Babylon interfaces used by FP are not changed between Babylon phase-2/3. The e2e test also passes

clientcontroller/babylon/babylon.go
gitferry: Good idea

gitferry:
> I think so, as the Babylon interfaces used by FP are not changed between Babylon phase-2/3. The e2e test also passes

> Nevertheless, I'm also concerned that this merge might mess up with the FP for phase-2, and would require another round of security audit. In this case @KonradStaniec Should we release FP for phase-2 like what we do in Babylon node?

We probably should do this

Lazar955 (approved):
Amazing work! Let's merge other PR's into this

SebastianElvis:
> * create long running `release/v1.x` branch before merging this pr to main
> * from `release/v1.x` branch cut the release of `v1.0.0-rc.0` finality provider
> * later progress with backports from main as in Babylon repo.

Agree with this plan. Will wait for more review and the v1 release before merging this PR.

KonradStaniec (approved):

PR #65, opened 2024-09-24
chore(e2e): babylond in docker

Lazar955:
Run e2e tests with babylond in docker container.

Seems that in order to execute tests with `t.Parallel()` we will need to do some refactoring in eots manager. Will be done in next PR.


gitferry (approved):
Great work!

PR #71, opened 2024-09-26
doc: Add core logic of finality provider and status transition

gitferry:


docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
+# Finality Provider Core
+
+This doc is to provide basic invariants and heuristics one should follow to
+build the finality provider. This part should be agnostic to any consumer
+chains.
+
+## Status Transitions
+
+The lifecycle of a finality provider instance involves 5 status: `Registered`,
+`Inactive`, `Active`, `Jailed`, `Slashed`. The status transition is depicted
```
SebastianElvis, line 10:
```suggestion
A finality provider has the following 5 possible states: `Registered`,
`Inactive`, `Active`, `Jailed`, `Slashed`. The state transition is depicted
```

docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
...
+in the following diagram.
+
+![Finality Provider Status Transition](./fp-status-transition.png).
+
+## Invariants
+
+We define invariants that the finality provider should satisfy:
+
+- Finality signature should be submitted for a block if:
```
SebastianElvis, line 19:
```suggestion
- The finality provider should submit a finality signature for a block if:
```

docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
...
+  - the block is not finalized,
+  - the finality provider has positive voting power for this block,
+  - the finality provider has not cast a finality signature for this block, and
+  - the finality provider has cast finality votes for all non-finalized blocks
+    prior to this block for which it has voting power.
+
+- Public randomness should be committed for a block if:
```
SebastianElvis, line 26:
```suggestion
- The finality provider should commit public randomness for a block if:
```

docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
...
+  - the block is not finalized, and
```
SebastianElvis, line 27: Is this constraint necessary and did we implement it?

docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
...
+  - the finality provider has not committed any randomness for this block
+
+- A finality provider instance should not be running if it’s status is `slashed`
+  or `jailed` on Babylon.
+
+#### Internal variables
```
SebastianElvis, line 33:
```suggestion
### Internal variables
```

docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
...
+  - the finality provider has not cast a finality signature for this block, and
+  - the finality provider has cast finality votes for all non-finalized blocks
+    prior to this block for which it has voting power.
+
+- Public randomness should be committed for a block if:
+  - the block is not finalized, and
+  - the finality provider has not committed any randomness for this block
+
+- A finality provider instance should not be running if it’s status is `slashed`
+  or `jailed` on Babylon.
```
SebastianElvis, line 31: Is it that the FP should not be running or this FP is still running but somehow remains silent?

docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
...
+
+#### Internal variables
+
+The finality provider maintains the following internal variables.
+* `last_voted_height`: the last block height on which the fp has cast a vote,
+* `last_processed_height`: the last block height on which the fp either has
+  cast a vote or it does not have voting power, and
+* `next_height`: the next block to poll.
```
SebastianElvis, line 39: Do you think we should distinguish between internal variables of the FP instance and the poller?

docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
...
+
+Internal variables need to satisfy the following rules:
+* `last_processed_height >= last_voted_height` should always hold, and
+* `next_height > max{latest_finalized_height, last_processed_height}` should
+  always hold.
```
SebastianElvis, line 44: Is 
```suggestion
* `next_height = max{latest_finalized_height, last_processed_height} + 1` should
  always hold.
```

always true?

docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Heuristics
+
+### Sending finality votes
+
+* The poller polls blocks from start_height one by one monotonically.
```
SebastianElvis, line 50:
```suggestion
* The poller polls blocks from `start_height` one by one monotonically.
```

docs/fp-core.md
```diff
@@ -0,0 +1,90 @@
...
+the waste is negligible.
+
+### Fast sync
+
+Fast sync is needed for bootstrap and in case the finality provider falls
+behind due to network latency.
+
+* The finality provider enters fast sync if:
+  * the finality provider is bootstrapping, or
+  * the `current_block_height - the last processed height > gap` (configured value).
```
SebastianElvis, line 80: After https://github.com/babylonlabs-io/finality-provider/pull/57, FP in the normal-case also batch submits finalkity signatures. Does it mean that the normal-case execution is same as fast sync with smaller gaps? If yes, does it make sense to harmoise the normal-case execution and fastsync, or even remove the fast sync completely?

SebastianElvis (commented):
Great doc, thanks! This clarifies a lot of things :+1:

docs/fp-core.md
gitferry: This indicates that if the block is already finalized, we can skip it but it does not mean that we cannot commit pub rand for a finalized block. We have implemented in https://github.com/babylonlabs-io/finality-provider/blob/8c63dae68284cb5426cb9cc07967e367c7d627cd/finality-provider/service/fp_instance.go#L502

docs/fp-core.md
gitferry: we have a monitor to check status of the running fp. if the status is detected to be `jailed` and `slashed`, we should stop the instance as it will be a waste of resources to run it. The operator should identify and fix the issue immediately. See our current impl https://github.com/babylonlabs-io/finality-provider/blob/8c63dae68284cb5426cb9cc07967e367c7d627cd/finality-provider/service/fp_manager.go#L130

docs/fp-core.md
gitferry: As there are not many internal variables to maintain, I think we can just list all of them here.

docs/fp-core.md
gitferry: Good point

docs/fp-core.md
gitferry: Possibly we can remove the case where the fp enters fast-sync from normal-execution but we still need fast sync for bootstrap. I'll keep it for now

SebastianElvis: I see, sg

SebastianElvis (approved):
lgtm!

PR #74, opened 2024-09-27
chore: use 0.5.0 rc version of Babylon

SebastianElvis:
After this PR, we can tag `euphrates-0.5.0-rc.0` for FP


maurolacy (commented):
LGTM.

PR #78, opened 2024-09-30
fix: fpd panics on jailed finality provider

gitferry:
In the previous implementation, once a finality provider is jailed, when the fpd is restarted, it will panic due to error from Babylon when fast sync. This PR fixed this issue with the following changes:
1. We implemented `unjail` finality provider CLI other than using the one from Babylon. This implementation will set the fp instance status to `inactive` after `unjail` tx is successfully sent (will later updated to active if the fp has voting power).
2. Once `jailed` error is detected while starting a fp instance, it will fail but the fpd will not panic, meaning that the loop for updating stored fp status continues running.

Now the flow of jailing/unjailing becomes the follows:
1. fpd detects jailing via err when sending a fp sig or a loop for checking fp status
2. once `jailed` detected, fpd terminates the fp instance without terminating the program
3. the operator checks fp signing info to get the `jail_until` via `babylond q finality signing-info [fp-pk-hex]`
4. after the `jail_until` is passed, the operator can unjail the fp by executing `fpd unjail-finality-provider [fp-pk-hex]`
5. if everything goes well, the fp will continue sending finality votes if it has voting power after a period of waiting for state transition


finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -174,6 +174,44 @@ func runCommandCreateFP(ctx client.Context, cmd *cobra.Command, _ []string) erro
 	return nil
 }
 
+// CommandUnjailFP returns the unjail-finality-provider command by connecting to the fpd daemon.
+func CommandUnjailFP() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:     "unjail-finality-provider",
+		Aliases: []string{"ufp"},
+		Short:   "Unjail the given finality provider.",
+		Example: fmt.Sprintf(`fpd unjail-finality-provider [eots_pk]--daemon-address %s ...`, defaultFpdDaemonAddress),
```
RafilxTenfen, line 183:
```suggestion
		Example: fmt.Sprintf(`fpd unjail-finality-provider [eots_pk] --daemon-address %s ...`, defaultFpdDaemonAddress),
```

finality-provider/cmd/fpd/daemon/tx.go
```diff
@@ -28,7 +27,6 @@ func CommandTxs() *cobra.Command {
 	cmd.AddCommand(
 		authcli.GetSignCommand(),
 		btcstakingcli.NewCreateFinalityProviderCmd(),
-		finalitycli.NewUnjailFinalityProviderCmd(),
```
RafilxTenfen, line 31: did you remove this tx to be able to run an update on the local status after the tx is run onchain?

gitferry: Exactly

finality-provider/service/fp_manager.go
```diff
@@ -247,6 +246,11 @@ func (fpm *FinalityProviderManager) StartFinalityProvider(fpPk *bbntypes.BIP340P
 	}
 
 	if err := fpm.addFinalityProviderInstance(fpPk, passphrase); err != nil {
+		if errors.Is(err, ErrFinalityProviderJailed) {
+			fpm.logger.Error("failed to start finality provider", zap.Error(err))
+			// do not return error
```
RafilxTenfen, line 251: why not? :thinking:

RafilxTenfen (approved):
utACK, only one question

finality-provider/service/fp_instance.go
```diff
@@ -98,10 +97,17 @@ func (fp *FinalityProviderInstance) Start() error {
...
+		return fmt.Errorf("%w: %s", ErrFinalityProviderJailed, fp.GetBtcPkHex())
+	}
+
 	fp.logger.Info("Starting finality-provider instance", zap.String("pk", fp.GetBtcPkHex()))
 
 	startHeight, err := fp.bootstrap()
 	if err != nil {
+		if errors.Is(err, ErrFinalityProviderJailed) {
+			fp.MustSetStatus(proto.FinalityProviderStatus_JAILED)
+		}
```
SebastianElvis, line 110: Do you think this should be done here or inside `bootstrap()`?

finality-provider/service/fp_manager.go
SebastianElvis: Same question here. This seems like an error to me

clientcontroller/babylon.go
```diff
@@ -257,6 +257,27 @@ func (bc *BabylonController) SubmitBatchFinalitySigs(
 	return &types.TxResponse{TxHash: res.TxHash, Events: res.Events}, nil
 }
 
+// UnjailFinalityProvider sends an unjail transaction to the consumer chain
+func (bc *BabylonController) UnjailFinalityProvider(fpPk *btcec.PublicKey) (*types.TxResponse, error) {
+	msg := &finalitytypes.MsgUnjailFinalityProvider{
+		Signer:  bc.mustGetTxSigner(),
+		FpBtcPk: bbntypes.NewBIP340PubKeyFromBTCPK(fpPk),
+	}
```
SebastianElvis, line 265: Do you think if it's beneficial to check whether the `jail_until` is passed before submitting this tx? If the operator submits unjail tx when the FP hasn't passed the jailing period, the tx will still consume gas but the FP won't get unjailed, which is not expected

SebastianElvis (approved):

clientcontroller/babylon.go
gitferry: Good point. I thought gas would not be confused as it will be rejected in the mempool level? Created an issue #79 to track this as this needs introduction of signing info queries and will be handled in a separate pr

finality-provider/service/fp_manager.go
gitferry: The point of not returning error is that we still want to start the app even if we fail to start the fp instance due to fp jailed. But I agree that we should not handle the error here. I changed to handle the error in the application level in [0eb53b3](https://github.com/babylonlabs-io/finality-provider/pull/78/commits/0eb53b366b0743a9334004d88901751e8a6d215e)

finality-provider/service/fp_instance.go
gitferry: Good point. Moved the error handler to `bootstrap()`

PR #80, opened 2024-10-01
backport: fpd panics on jailed finality provider (#78)

gitferry:


SebastianElvis:
Does every new commit to `main` need to be backported to the release branch? I was under the impression that 

- release branches will only accept hotfixes
- hotfixes on the release branches need to be backported to `main`
- new features on `main` shall be released in the next major version

imo this seems to be a feature and will need to go to the release. Or did I miss anything?

gitferry:
@SebastianElvis yep, I agree with the above. This PR is a hotfix and should be backported to the release branch as it will be deployed for the current devnet

SebastianElvis (approved):
Sounds good, approving then. 

Btw, for such hotfix, should it be pushed to `main` first or to `release/v0.5.x` first? I thought it should be the latter

gitferry:
> Btw, for such hotfix, should it be pushed to main first or to release/v0.5.x first? I thought it should be the latter

Agreed

bap2pecs:
> This approach ensures that main has always al relevent fixes.

i guess this also applies to not just hotfixes. anything that can easily apply to main (e.g. CI, refactor, cleanup, fix race condition) can target main first and backport w a separate PR

PR #83, opened 2024-10-04
chore: Enable gosec and fix findings

vitsalis:
This update involves:
- Enabling gosec in the github workflow and updating the github workflow versions.
- Update the codebase to fix findings from gosec. Highlights:
   - Some unhandled errors
   - Config types allowed for very large values. For example, the retry intervals and the commit gaps allowed for uint64. All values related to block sizes were left to uint64, but the others were reduced to uint32.
   - Prometheus server protection against Slowloris

### DIFF
--- a/metrics/server.go
+++ b/metrics/server.go
@@ -3,6 +3,7 @@ package metrics
 import (
 	"context"
 	"net/http"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"go.uber.org/zap"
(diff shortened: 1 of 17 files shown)


gitferry (approved):
Thanks for adding gosec!

metrics/server.go
```diff
@@ -20,8 +21,12 @@ func Start(addr string, logger *zap.Logger) *Server {
 
 	// Create the HTTP server with the custom ServeMux as the handler
 	server := &http.Server{
-		Addr:    addr,
-		Handler: mux,
+		Addr:              addr,
+		Handler:           mux,
+		ReadHeaderTimeout: 2 * time.Second,
+		ReadTimeout:       1 * time.Second,
```
KonradStaniec, line 27: I would say that timeouts for reading are bit to aggressive here. 

I would set both to `5seconds`. Mostly becouse I saw such values used in:
- https://blog.cloudflare.com/exposing-go-on-the-internet/
- and https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

KonradStaniec (approved):
looks good, just small comments about timeouts

metrics/server.go
vitsalis: done

filippos47 (approved):
Looking good, thank you!

Issue #63, opened 2024-09-24
Improve eotsd key management


gitferry:
* Passphrase is passed to the daemon via CLI flag in `keys add`, `export-pop`, and `sign-schnorr`, which is unsafe

PR #86, opened 2024-10-08
chore: Remove running multiple fp instances support

gitferry:
This PR partly addresses https://github.com/babylonlabs-io/finality-provider/issues/34 by disallowing running multiple fp instances within the fp manager. Keyring level improvement will be done in a separate PR to address #63. Notable changes in this PR:

1. Remove `MaxNumFinalityProviders` from config as it is not applicable
2. If `--eots-pk` is not specified in `fpd start`, fp app will start without any fp instances. To run a fp instance, `--eots-pk` must be given
3. fp manager only keeps a single fp instance instead of a map

### DIFF
--- a/finality-provider/cmd/fpd/daemon/start.go
+++ b/finality-provider/cmd/fpd/daemon/start.go
@@ -120,35 +120,31 @@ func startApp(
 	fpApp *service.FinalityProviderApp,
 	fpPkStr, passphrase string,
 ) error {
-	// only start the app without starting any finality-provider instance
-	// as there might be no finality-provider registered yet
+	// only start the app without starting any finality provider instance
+	// this is needed for new finality provider registration or unjailing
+	// finality providers
 	if err := fpApp.Start(); err != nil {
-		return fmt.Errorf("failed to start the finality-provider app: %w", err)
+		return fmt.Errorf("failed to start the finality provider app: %w", err)
 	}
 
-	if fpPkStr != "" {
-		// start the finality-provider instance with the given public key
-		fpPk, err := types.NewBIP340PubKeyFromHex(fpPkStr)
-		if err != nil {
-			return fmt.Errorf("invalid finality-provider public key %s: %w", fpPkStr, err)
-		}
-
-		if err := fpApp.StartHandlingFinalityProvider(fpPk, passphrase); err != nil {
-			if errors.Is(err, service.ErrFinalityProviderJailed) {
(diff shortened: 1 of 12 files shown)


Lazar955 (approved):

RafilxTenfen (approved):
LGTM

finality-provider/service/app.go
SebastianElvis, line 1: Do we have to check whether FP is registered upon handling `createFinalityProviderRequestChan`?

finality-provider/service/fp_manager.go
```diff
@@ -31,15 +30,15 @@ func (ce *CriticalError) Error() string {
 	return fmt.Sprintf("critical err on finality-provider %s: %s", ce.fpBtcPk.MarshalHex(), ce.err.Error())
 }
 
+// FinalityProviderManager is responsible to initiate and start the given finality
+// provider instance, monitor its running status
 type FinalityProviderManager struct {
```
SebastianElvis, line 35: Have we considered removing one of FP manager or FP instance completely? The manager-instance separation is mostly due to the multi FP feature but now we no longer need it

finality-provider/cmd/fpd/daemon/start.go
```diff
@@ -120,35 +120,31 @@ func startApp(
...
-	if fpPkStr != "" {
-		// start the finality-provider instance with the given public key
-		fpPk, err := types.NewBIP340PubKeyFromHex(fpPkStr)
-		if err != nil {
-			return fmt.Errorf("invalid finality-provider public key %s: %w", fpPkStr, err)
-		}
+	// no fp instance will be started if public key is not specified
+	if fpPkStr == "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
gitferry: If no pk is specified, we do not need to start any instance. And this is not an error. This is needed when we don't have any fp in the store yet, so we just start the application and then use daemon commands to create a new fp

finality-provider/service/app.go
gitferry: Probably not. In this stage, we only create fp in the local store. If the fp is already registered, then we will get error when we try to register it to Babylon

finality-provider/service/fp_manager.go
gitferry: I agree we probably don't need the manager and the goroutines in it can be moved to the instance itself. Will do it in a separate PR as it will touch more things

gitferry:
Merging this PR for now. Will address the remaining comments in a separate PR

PR #90, opened 2024-10-10
feat(cmd): edit finality provider description

Lazar955:
- Adds CLI support to edit finality provider description

_Note: I did try using the edit cmd defined in Babylon, but I wasn't successful as the cmd context is getting passed into the edit cmd resulting in a flags error._

[References issue](https://github.com/babylonlabs-io/finality-provider/issues/12)


RafilxTenfen:
Could you share the flag error it was giving?

Lazar955:
> Could you share the flag error it was giving?

@RafilxTenfen it's inheriting the flags passed, so pass in the "daemon-address" flag, which also gets passed to the Babylon command, resulting in unexpected flag. 

But in anyway, I find this approach better, as you are not handling things on the "server" rather than on the "client". As of now CLI just calls the RPC where everything is handled. In the other approach where the CLI would:
1. Fetch the FP from bbn
2. Determine to set or not flags for the  bbn edit cmd
3. Invoke the cmd on bbn
4. Still invoke the FP RPC, because we need to update the local store

finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -446,6 +446,71 @@ func runCommandAddFinalitySig(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// CommandEditFinalityDescription edits description of finality provider
+func CommandEditFinalityDescription() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:     "edit-finality-provider [btc_pk]",
+		Aliases: []string{"efp"},
+		Short:   "Edit finality provider",
+		Example: fmt.Sprintf(`fpd edit-finality-provider --daemon-address %s`, defaultFpdDaemonAddress),
```
gitferry, line 455: maybe it's better to provide an example including one of the editable fields, e.g., moniker

finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -446,6 +446,71 @@ func runCommandAddFinalitySig(cmd *cobra.Command, args []string) error {
...
+	desc := &proto.Description{
+		Moniker:         moniker,
+		Identity:        identity,
+		Website:         website,
+		SecurityContact: securityContact,
+		Details:         details,
+	}
+
+	if err := grpcClient.EditFinalityProvider(cmd.Context(), fpPk, desc); err != nil {
+		return fmt.Errorf("failed to get finality provider %v err %v", fpPk.MarshalHex(), err)
```
gitferry, line 506:
```suggestion
		return fmt.Errorf("failed to edit finality provider %v err %v", fpPk.MarshalHex(), err)
```

finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -446,6 +446,71 @@ func runCommandAddFinalitySig(cmd *cobra.Command, args []string) error {
...
+
+	desc := &proto.Description{
+		Moniker:         moniker,
+		Identity:        identity,
+		Website:         website,
+		SecurityContact: securityContact,
+		Details:         details,
+	}
+
+	if err := grpcClient.EditFinalityProvider(cmd.Context(), fpPk, desc); err != nil {
```
gitferry, line 505: seems we are missing the commission rate here? We should also allow users to change the commission rate

clientcontroller/babylon.go
```diff
@@ -510,6 +510,66 @@ func (bc *BabylonController) QueryFinalityProviders() ([]*btcstakingtypes.Finali
...
+	fpPubKey := bbntypes.NewBIP340PubKeyFromBTCPK(fpPk)
+	res, err := bc.bbnClient.QueryClient.FinalityProvider(fpPubKey.MarshalHex())
+	if err != nil {
+		return nil, fmt.Errorf("failed to query the finality provider %s: %v", fpPubKey.MarshalHex(), err)
+	}
+
+	return res, nil
+}
+
+func (bc *BabylonController) EditFinalityProviderDescription(fpPk *btcec.PublicKey,
```
gitferry, line 523: It's better to have a docstring to explain that default value from Babylon will be used if any descriptive fields are not updated

clientcontroller/interface.go
```diff
@@ -49,6 +50,9 @@ type ClientController interface {
 	// QueryFinalityProviderSlashedOrJailed queries if the finality provider is slashed or jailed
 	QueryFinalityProviderSlashedOrJailed(fpPk *btcec.PublicKey) (slashed bool, jailed bool, err error)
 
+	// EditFinalityProviderDescription edits description of the finality provider
+	EditFinalityProviderDescription(fpPk *btcec.PublicKey, reqDesc sttypes.Description) (*sttypes.Description, error)
```
gitferry, line 54: It would be more general if we could have our own description type as our consumer chains might not be a Cosmos chain

clientcontroller/babylon.go
```diff
@@ -510,6 +510,66 @@ func (bc *BabylonController) QueryFinalityProviders() ([]*btcstakingtypes.Finali
...
+	reqDesc sttypes.Description) (*sttypes.Description, error) {
+
+	fpRes, err := bc.QueryFinalityProvider(fpPk)
```
gitferry, line 526: I think we should check `fpRes.Addr == bc.mustGetTxSigner()` as we require that only the one that controlls the babylon key of the finality provider can edit the fp

finality-provider/cmd/fpd/daemon/daemon_commands.go
Lazar955: Got it, thought we were only editing description

clientcontroller/babylon.go
Lazar955: Make sense, although this is checked on bbn, make sense to validate it here also to prevent unnecessary RPC

RafilxTenfen (approved):
In my mind it would be easier to take advantage of the CLI that already exists on Babylon like #92 and add a [post run](https://github.com/babylonlabs-io/finality-provider/blob/7a0311b46e299aa32a02637769ea12859b9c8220/finality-provider/cmd/fpd/daemon/keys.go#L25) to update the local storage of fpd if needed 

But, since this is already developed and it does something more than just call the transaction (set default values) the code looks good to me

gitferry (approved):
Lgtm! Some minor comments:

gitferry:
Seems CI was not triggered?

clientcontroller/interface.go
```diff
@@ -49,6 +51,9 @@ type ClientController interface {
 	// QueryFinalityProviderSlashedOrJailed queries if the finality provider is slashed or jailed
 	QueryFinalityProviderSlashedOrJailed(fpPk *btcec.PublicKey) (slashed bool, jailed bool, err error)
 
+	// EditFinalityProvider edits description and commission of a finality provider
+	EditFinalityProvider(fpPk *btcec.PublicKey, rate *sdkmath.LegacyDec, reqDesc *proto.Description) (*btcstakingtypes.MsgEditFinalityProvider, error)
```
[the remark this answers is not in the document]
Lazar955: Ah, agreed, good point

Lazar955:
> Seems CI was not triggered?

prob because of conflict with the changelog

PR #91, opened 2024-10-11
ci: add goreleaser setup and move out changelog reminder

SebastianElvis:
This PR

- sets up Go releaser for Babylon. Currently it only supports linux amd64. Other targets will be supported in the future. To test locally, run `make release-snapshot`
- moves changelog reminder out from the CI flow


SebastianElvis:
> Nice work! Couldn't tests as I'm on mac

Probably you could give it a try as the releaser is using a Docker image that is supposed to be cross-platform?

PR #96, opened 2024-10-16
nit: fix workflow syntax issue

bap2pecs:
not sure why it didn't detect in https://github.com/babylonlabs-io/finality-provider/pull/95

image

I have no idea why it's failing but I noticed this formatting issue


bap2pecs:
image

unit test seems flaky. encounter this problem a few times lately. i am rerunning the job

PR #101, opened 2024-10-19
chore: add finality activation height

RafilxTenfen:
- Add checks for finality activation height
- Pub rand commit start height at activation height or last commited height whatever is higher
- Finality vote, only vote in blocks after the finality activation height
- Update babylon to v014


itest/test_manager.go
```diff
@@ -905,13 +906,13 @@ func ParseRespBTCDelToBTCDel(resp *bstypes.BTCDelegationResponse) (btcDel *bstyp
...
 		}
 
-		if len(ud.DelegatorUnbondingSigHex) > 0 {
-			delUnbondingSig, err := bbntypes.NewBIP340SignatureFromHex(ud.DelegatorUnbondingSigHex)
-			if err != nil {
-				return nil, err
-			}
-			btcDel.BtcUndelegation.DelegatorUnbondingSig = delUnbondingSig
-		}
+		// if len(ud.DelegatorSlashingSigHex.) > 0 {
```
KonradStaniec, line 909: leftover ?

KonradStaniec (approved):
In general looks good 👍 though lets wait for @gitferry review

itest/test_manager.go
RafilxTenfen: yes, removing it 
Thanks

finality-provider/service/fastsync.go
```diff
@@ -53,6 +58,10 @@ func (fp *FinalityProviderInstance) FastSync(startHeight, endHeight uint64) (*Fa
 			if fp.hasProcessed(b) {
 				continue
 			}
+			// check if it is allowed to send finality
+			if b.Height < activationBlkHeight {
+				continue
+			}
```
gitferry, line 64: maybe this check can be moved out of the for loop by comparing the activation height against start height and end height?

finality-provider/service/fp_instance.go
```diff
@@ -395,6 +389,18 @@ func (fp *FinalityProviderInstance) tryFastSync(targetBlock *types.BlockInfo) (*
 	return fp.FastSync(startHeight, targetBlock.Height)
 }
 
+func (fp *FinalityProviderInstance) fastSyncStartHeight(lastFinalizedHeight uint64) (uint64, error) {
```
gitferry, line 392: `fastSyncStartHeight` is called after we have known that there's finalized blocks, indicating that activation is already started, so maybe we don't need to check activation height here?

util/math.go
```diff
@@ -0,0 +1,11 @@
+package util
+
+func MaxUint64(values ...uint64) (max uint64) {
```
Lazar955, line 3: Nit: 
1. Could be a generic max function
2. I find it to be an antipattern to have a `util` package (becomes a code dumpster):
- suggestion: `lib/math/math.go`

Lazar955 (approved):
Lgtm! Minor nits:

finality-provider/service/fastsync.go
RafilxTenfen: done at https://github.com/babylonlabs-io/finality-provider/pull/101/commits/e03994d40a2c50373df6793f63c1be78e4e0ec3f

finality-provider/service/fp_instance.go
RafilxTenfen: fastSyncStartHeight returns the height that fast sync should start right? 
I probably can remove the check inside the fastSync than

RafilxTenfen:
> Maybe I was missing something, but shouldn't we add a check to prevent the fp sending finality sig before activation height?

Yeap, you are right, there was another loop that sends finality, check added at https://github.com/babylonlabs-io/finality-provider/pull/101/commits/7175eba9f6a54b7fc4b1d5273ea3f09ee687eb6a

gitferry (approved):
Thanks for the changes! Lgtm!

util/math.go
RafilxTenfen: agreed, moved 
generic did not worked that easy, so I kept as MaxUint64

PR #102, opened 2024-10-22
chore: refactor eotsd keys cmd

### LINKED ISSUE #63: Improve eotsd key management

gitferry:
Closes #63:
* Replace `eotsd` cli library with `cobra` lib to be consistent with `fpd`
* Remove `export-pop` and `sign-schnorr` because they are not needed for phase-2
* Reuse `keys` commands from `cosmos-sdk`
* Wrap `keys add` commands to save the public key and key name wrapping
* Refactor eotsd cmd folder architect

Example after change:
```
❯ ./build/eotsd keys add "test-key-0" --keyring-backend=test

**Important** write this mnemonic phrase in a safe place.
It is the only way to recover your account if you ever forget your password.

topic expire slush beyond room cube train tenant dice negative winter lamp unaware winner story bounce space try rotate tiny assault grain scheme pact

name: test-key-0
pub_key_hex: e40ea1364293afd9d7400b58f336e7b6b10baf83ef31ef83b447f3fcdd928b59
```


eotsmanager/cmd/eotsd/utils.go
```diff
@@ -0,0 +1,58 @@
...
+
+	return homePath, nil
+}
+
+// PersistClientCtx persist some vars from the cmd or config to the client context.
+// It gives preferences to flags over the values in the config. If the flag is not set
+// and exists a value in the config that could be used, it will be set in the ctx.
+func PersistClientCtx(ctx client.Context) func(cmd *cobra.Command, _ []string) error {
+	return func(cmd *cobra.Command, _ []string) error {
+		// TODO(verify): if it uses the default encoding config it fails to list keys! output:
```
KonradStaniec, line 36: oh do we have some test case for this ? or this is cosmos-sdk bug and we do not need to do anything with this ?

KonradStaniec (approved):
just one question about this `todo`

eotsmanager/cmd/eotsd/utils.go
gitferry: Oh I copied it from https://github.com/babylonlabs-io/finality-provider/blob/f7c17b17e6ec546ac9370b9a3d21686eeb5e12a7/finality-provider/cmd/cmd.go#L18. Any ideas? @RafilxTenfen

RafilxTenfen: Oh, yeah I remembered 
I only create the whole app to get the configs to ctx

```go
		tempApp := app.NewTmpBabylonApp()

		ctx = ctx.
			WithCodec(tempApp.AppCodec()).
			WithInterfaceRegistry(tempApp.InterfaceRegistry()).
			WithTxConfig(tempApp.TxConfig()).
			WithLegacyAmino(tempApp.LegacyAmino()).
			WithInput(os.Stdin)
```

If we use the `appparams.DefaultEncodingConfig()` instead of `tempApp := app.NewTmpBabylonApp()` it fails

gitferry: I think we need to add `std.RegisterInterfaces(encCfg.InterfaceRegistry)`. Otherwise, it does not how to decode the key. See the fix in [942df6e](https://github.com/babylonlabs-io/finality-provider/pull/102/commits/942df6e2ec3e1c391156effa704d44feec7739e2)

PR #104, opened 2024-10-24
feat(cmd): print binary version in fpd

Lazar955:
Running `fpd version` now prints info on the built binary:

image

[Related to issue](https://github.com/babylonlabs-io/finality-provider/issues/11)


gitferry (approved):
Excellent!

PR #117, opened 2024-11-06
doc: Spec of commit public randomness

gitferry:


Lazar955 (approved):
Nice write-up!

docs/commit-pub-rand.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Overview
+
+The finality provider periodically commits public randomness to the consumer
+chain to be used for future block finalization. This document specifies the
+process of committing public randomness.
+
+## Commit Process
+
+### Generating a Commit
```
KonradStaniec, line 11: I think I would describe here also:
- full commit message used in Babylon (can be link to Babylon proto file)
- the format of randomness i.e 32byte public point
- the merkle tree library used and format of the proof

In general, this spec should have enough details that any body can implement his own randomness committing process.

docs/commit-pub-rand.md
```diff
@@ -0,0 +1,124 @@
...
+To generate a new commit, following steps are needed:
+
+- Generate a list of randomness. This requires a RPC call to the EOTS manager
+  (eotsd) to generate a list of public randomness, each corresponding to a
+  specific height according to the start height and the number of randomness in
+  the request.
+- Construct the merkle tree based on the list of randomness and save the merkle
+  proof of each randomness to the database indexed by height.
+- Construct the commit message based on the root of the merkle tree.
+- Sign the commit message and sent a transaction to Babylon.
```
KonradStaniec, line 29: I would describe here what data is signed and using which signature algo

docs/commit-pub-rand.md
gitferry: Good point! Added useful links in [a73f745](https://github.com/babylonlabs-io/finality-provider/pull/117/commits/a73f7457cf1fd2f845935c7ed80b6351b2900209)

KonradStaniec (approved):
Just a small comment about the fact that spec should be a bit more general, but other than that it looks good.

docs/commit-pub-rand.md
```diff
@@ -0,0 +1,132 @@
...
+
+To generate a new commit, following steps are needed:
+
+- Generate a list of randomness. This requires an RPC call to the EOTS manager
+  (eotsd) to generate a list of public randomness, each corresponding to a
+  specific height according to the start height and the number of randomness in
+  the request. The details of the generator can be found in [eotsmanager/randgenerator](../eotsmanager/randgenerator/randgenerator.go).
+- Construct the merkle tree based on the list of randomness using the CometBFT's [merkle](https://github.com/cometbft/cometbft/tree/main/crypto/merkle)
+  library.
+- Save each randomness along with the merkle
```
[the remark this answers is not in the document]
gitferry: Makes sense. Improved at [ac29934](https://github.com/babylonlabs-io/finality-provider/pull/117/commits/ac29934f95e449415ba71dfbb3f8d8303750c16e)

PR #120, opened 2024-11-07
doc: Spec of finality vote submission

gitferry:


docs/send-finality-vote.md
```diff
@@ -0,0 +1,82 @@
...
+
+1. Requests a finality signature from the EOTS manager
+2. Submits the vote transaction to the consumer chain
+3. Implements retry logic until either:
+   - Maximum retry attempts are reached
+   - Block becomes finalized
+
+#### Batch submission
+
+A batch submission mechanism is needed to deal with cases where:
```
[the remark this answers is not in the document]
gitferry: In the current code, all the submission is done in the submission loop. So there's no race condition. The fast sync loop is for sending the lagging signal. If lagging is detected, the sumission loop will do fast sync. See https://github.com/babylonlabs-io/finality-provider/blob/3f8aa013a72db9b8b9cb367a0af1155c3decd413/finality-provider/service/fp_instance.go#L254.

Nevertheless, in the spec, we use batch submission other than fast sync and we plan to remove the latter completely.

docs/send-finality-vote.md
```diff
@@ -0,0 +1,82 @@
+# Finality Votes Submission Specification
+
+## Overview
+
+Finality providers submit votes to finalize blocks on the consumer chain.
+This document specifies the process of submitting finality votes.
+
+## Internal State
+
+The finality provider maintains two critical persistent states:
```
bap2pecs, line 10: i think it's simpler to just keep one internal state i.e. last processed height

last voted height can be easily retrieved from remote

docs/send-finality-vote.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Submission Process
+
+### Bootstraping
+
+To determine the initial processing height:
+
+1. Query consumer chain for `lastFinalizedHeight`
+2. If no finalized blocks exist:
+   - Start from `finalityActivationHeight`
```
bap2pecs, line 34: there is a problem here.

imagine there is no finalized block, but this FP already voted for the first activated block and it's just waiting for the rest FP to vote so the block can get finalized

starting at `finalityActivationHeight` means a duplicate vote

docs/send-finality-vote.md
bap2pecs: actually do you think we can get rid of both from DB and only keeps one in-memory state (i.e. `fp.nextHeightToProcess`)?

I feel it's doable

docs/send-finality-vote.md
```diff
@@ -0,0 +1,82 @@
...
+
+- Last voted height
+- Last processed height
+
+### Last voted height
+
+Tracks the most recent height for which a finality vote was successfully
+submitted. This prevents duplicate voting on previously voted heights.
+
+### Last processed height
```
parketh, line 20: what's the difference between last voted height and last processed height?

docs/send-finality-vote.md
gitferry: Good point. I think we maybe can get rid of `lastVotedHeight` but why do you say it's easy to retrieve `lastVotedHeight` from the remote? It seems that we need to save this info in the finality provider on Babylon

> actually do you think we can get rid of both from DB and only keeps one in-memory state (i.e. fp.nextHeightToProcess)?

if we get rid of `fp.nextHeightToProcess`, we will waste some resources processing blocks that have already been processed, won't we?

docs/send-finality-vote.md
gitferry: Good point. See updates in [ab8913a](https://github.com/babylonlabs-io/finality-provider/pull/120/commits/ab8913acdc3e03e7ad36aebad133075942916131)

docs/send-finality-vote.md
gitferry: last processed height also covers the case where the fp does not have voting power

docs/send-finality-vote.md
bap2pecs: > but why do you say it's easy to retrieve lastVotedHeight from the remote? It seems that we need to save this info in the finality provider on Babylon

b/c if we store it in a local DB, there will be two source of truth. for the OP consumer FP, we can easily add a query in the CosmWasm contract for it e.g. `getFpLastVotedHeight`

bap2pecs: > if we get rid of fp.nextHeightToProcess, we will waste some resources processing blocks that have already been processed, won't we?

no I mean to get rid of the local DB states but keep `nextHeightToProcess`

how it works is: at startup, the FP will get everything it needs from the Babylon Chain and/or the CosmWasm contract to set `nextHeightToProcess`

but `nextHeightToProcess` is not a DB state, it's an in-memory field value of the FinalityProvider object

gitferry: > b/c if we store it in a local DB, there will be two source of truth. for the OP consumer FP, we can easily add a query in the CosmWasm contract for it e.g. getFpLastVotedHeight

Does CosmWasm contract maintain `lastVotedHeight` for query? I don't think Babylon stores such information. I think it's worth adding it. Wdyt @SebastianElvis?

gitferry: > how it works is: at startup, the FP will get everything it needs from the Babylon Chain and/or the CosmWasm contract to set nextHeightToProcess

If I understand it correctly, we only have `lastVotedHeight` to update the in-memory field `nextHeightToProcess`, right? If the fp does not have voting power for a lot of blocks, after processing them and restart, the the fp needs to process those block again

bap2pecs: These are specific for Consumer. Each consumer chain deploys its own CW contract storing the data

bap2pecs: For Babylon FP, you can probably add a similar query API directly

bap2pecs: > If I understand it correctly, we only have `lastVotedHeight` to update the in-memory field `nextHeightToProcess`, right?

maybe I misunderstood, I thought you are proposing to have two DB states:

- Last voted height
- Last processed height

my proposal is we don't need neither to update the in-memory field `nextHeightToProcess`. we always read from remote at startup time

>  If the fp does not have voting power for a lot of blocks, after processing them and restart, the the fp needs to process those block again

right this is an edge case which I think won't happen too frequently. you need to satisfy: 1) FP doesn't have power; 2) You restart the FP when it doesn't have power. This doesn't see a frequent use case.

I'd vote instead to keep the architecture simple than optimize for it.

gitferry: > For Babylon FP, you can probably add a similar query API directly

I don't think so as `LastVotedHeight` is not stored anywhere, but I think it's worth storing it.

> my proposal is we don't need neither to update the in-memory field nextHeightToProcess. we always read from remote at startup time

I meant we can only read `lastVotedHeight` from the remote to update `nextHeightToProcess`

> right this is an edge case which I think won't happen too frequently. you need to satisfy: 1) FP doesn't have power; 2) You restart the FP when it doesn't have power. This doesn't seem like a frequent use case.

It's not a frequent case but it's a valid case and persisting `lastProcessedHeight` can address it easily. And, the inconsistency of `lastProcessedHeight` does not hurt security. So I would rather keep it as it is now unless you think it's not secure to do so

gitferry: Makes sense. Will bring it to the team to make the decision. Appreciate the discussion!

gitferry: Yep, I don't think it's safe to assume that FP is connected to a trusted RPC as we can not control it. The bottom line is that fp should not sign twice for the same height.

docs/send-finality-vote.md
```diff
@@ -0,0 +1,90 @@
...
+
+To determine the initial processing height:
+
+1. Query consumer chain for `lastFinalizedHeight` (`0` if no finalized blocks)
+2. Query consumer chain for `finalityActivationHeight`:
+3. Query `lastVotedHeightRemote` from consumer chain (`0` if no votes are sent)
+4. Synchronize local state:
+   - Ensure local `lastVotedHeight` and `lastProcessedHeight` ≥ `lastVotedHeightRemote`
+5. Begin processing at:
+   - `max(lastProcessedHeight + 1, lastFinalizedHeight + 1, finalityActivationHeight)`
```
KonradStaniec, line 38: two things:
- `lastProcessedHeight` is not initialised anywhere. I assume it comes from persistent state. Lets add the step of retrievieng it from database
- in some consumer chains, it maybe possible to vote and receive reward even after finality is reached, so starting from `lastFinalizedHeight + 1` may leave some rewards on the table. I would maybe leave it as comment here.

docs/send-finality-vote.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Normal Submission Loop
+
+After the finality provider is bootstraped, it continuously monitors for
+new blocks. For each new block, it performs these validation checks:
```
KonradStaniec, line 43: Do you think it is worth mentioning that finality provider receives block from trusted full node of consumer chain (or at least light client ) ?

bap2pecs:
how should we handle these two cases:
1. pub rand is lagging behind a lot to the tip height
2. FP is currently in jail or just get out of jail

docs/send-finality-vote.md
gitferry: > lastProcessedHeight is not initialised anywhere. I assume it comes from persistent state. Lets add the step of retrievieng it from database

In the discussion [here](https://github.com/babylonlabs-io/finality-provider/pull/120#discussion_r1836699698) I think we could probably be able to remove `lastProcessedHeight` as it is not critical to security.

> in some consumer chains, it maybe possible to vote and receive reward even after finality is reached, so starting from lastFinalizedHeight + 1 may leave some rewards on the table. I would maybe leave it as comment here.

Good point. Fixed in [8732f15](https://github.com/babylonlabs-io/finality-provider/pull/120/commits/8732f15c540d9f4361b9dace934cf4a87aebb3e6)

docs/send-finality-vote.md
gitferry: Good point. Fixed in [8732f15](https://github.com/babylonlabs-io/finality-provider/pull/120/commits/8732f15c540d9f4361b9dace934cf4a87aebb3e6)

gitferry:
> pub rand is lagging behind a lot to the tip height

When a fp is started, a sperate routine for committing public randomness should be started. This part if not the cope of this spec.

> FP is currently in jail or just get out of jail

In this case, the operator should get the fp unjailed first. Will have a separate doc for this

bap2pecs:
> When a fp is started, a sperate routine for committing public randomness should be started. This part if not the cope of this spec.

but the vote submission routine has to wait for it, so it's still relevant?

gitferry:
> but the vote submission routine has to wait for the PR is timestamped, so it's still relevant?

If the PR for a block is not timestamped, then the voting power is `0` for good. Even though the PR is timestamped later, the result won't change

bap2pecs:
> If the PR for a block is not timestamped, then the voting power is `0` for good

I am not aware of this. shouldn't vp only be determined by the btc delegations?

gitferry:
> I am not aware of this. shouldn't vp only be determined by the btc delegations?

This condition was added ffter integration with BTC-timestamping, see the [figure](https://github.com/babylonlabs-io/finality-provider/blob/608038c4b5b403b242c923c5f1c686617201b2b8/docs/fp-status-transition.png) here: a fp will only have voting power if (1) it has top N delegations, and (2) it has timestamped randomness for the block

bap2pecs:
> (1) it has top N delegations

what does this mean?

bap2pecs:
> This condition was added ffter integration with BTC-timestamping, see the [figure](https://github.com/babylonlabs-io/finality-provider/blob/608038c4b5b403b242c923c5f1c686617201b2b8/docs/fp-status-transition.png) here: a fp will only have voting power if (1) it has top N delegations, and (2) it has timestamped randomness for the block

this is good to know. but the important implication is when we have PR lagging for whatever reason (the 4 scenarios as I described in https://github.com/babylonlabs-io/finality-provider/pull/131), it's possible that the entire consumer chain's total VP becomes zero for some range of blocks. so FPs will skip voting for those blocks. I guess there is nothing much we can do in such an extreme case, but it is worth pointing out. cc @SebastianElvis

gitferry:
> this is good to know. but the important implication is when we have PR lagging for whatever reason (the 4 scenarios as I described in https://github.com/babylonlabs-io/finality-provider/pull/131), it's possible that the entire consumer chain's total VP becomes zero for some range of blocks. so FPs will skip voting for those blocks. I guess there is nothing much we can do in such an extreme case, but it is worth pointing out. cc @SebastianElvis

Yes, this is an extreme case but valid. In this case, we would have a period where no blocks will be finalized until at least one fp has voting power and sends the vote

KonradStaniec (approved):

PR #127, opened 2024-11-13
fix dockerfile issues

huynaism:


.github/workflows/publish.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - 'main'
       - 'base/consumer-chain-support'
+      - 'devops/add-docker-sec'
```
gitferry, line 8: seems can remove this?

gitferry (approved):
LGTM! Please also update the changelog

.github/workflows/publish.yml
huynaism: ok removed

Issue #115, opened 2024-11-05
Unable to recover due to lack of public randomness


gitferry:
After a long downtime of an fp, the committed public randomness could be run out. During bootstrapping, it should ensure the committed randomness is sufficient. This could lead to the following errors:

```
Whoops. There was an error while executing your fpd CLI 'failed to start app: failed to start finality-provider 5238b0223ea9d9d59ed4371e6248bfc55b14193f862c4773ac2972d70289541b instance: failed to bootstrap the finality-provider 5238b0223ea9d9d59ed4371e6248bfc55b14193f862c4773ac2972d70289541b: failed to send a batch of finality signatures to the consumer chain: rpc error: code = Unknown desc = rpc error: code = Unknown desc = failed to execute message; message index: 6: Public randomness not found for finality provider 5238b0223ea9d9d59ed4371e6248bfc55b14193f862c4773ac2972d70289541b at height 88301: execute wasm contract failed [CosmWasm/wasmd@v0.51.0/x/wasm/keeper/keeper.go:422] with gas used: '1728195': unknown request'%
```

```
Whoops. There was an error while executing your fpd CLI 'failed to start app: failed to start finality-provider dbaec7f5ccaf56a3b02d0af1ae1480623f6c1d132450c90e0da391ee88a3aa98 instance: failed to bootstrap the finality-provider dbaec7f5ccaf56a3b02d0af1ae1480623f6c1d132450c90e0da391ee88a3aa98: failed to get public randomness inclusion proof list: public randomness proof not found
```

bap2pecs:
any updates on this?

jfyi I am currently adding another daemon command `fpd command-pubrand <fpBtcPk> <targetHeight>` to help recover FP stuck in this case

cc @SebastianElvis @gitferry

bap2pecs:
What will the proper fix look like? I think we will need to get rid of fast sync in bootstrap() but the change is non-trivial

bap2pecs:
> jfyi I am currently adding another daemon command `fpd commit-pubrand <fpBtcPk> <targetHeight>` to help recover FP stuck in this case

https://github.com/babylonlabs-io/finality-provider/pull/131

gitferry:
Thanks for the reminder. The goal is to make sure sending votes and committing randomness are entirely independent routines (fp will not have voting power if no timestamped randomness). Will look into the solution this week

bap2pecs:
[...]

PR #131, opened 2024-11-18
[RFC] feat: add new cmd `fpd commit-pubrand [fp-eots-pk-hex] [block-height]`

bap2pecs:
## Summary

https://github.com/babylonlabs-io/finality-provider/issues/115

At `bootstrap()`, FP will fast sync to try to fast sync to the latest block. however, this makes a false assumption that PR is always ready.

So if the PR is lagging behind, fast sync will always fail b/c we never fast sync PRs. Then `bootstrap()` will stuck forever.

There are a few scenarios that can lead to this situation (we met with most of them):
- FP run out of BBN for a long time
- L2 RPC is down for a long time
- FP is down for a long time
- Babylon RPC is down for a long time

A proper fix is to remove fast sync from the `bootstrap()` step but that needs a massive refactor. As we discussed, we will use batch processing and get rid of fast sync completely.

For now, this helper command will help us recover FPs that are stuck at bootstrap.

## TODOs

I am sending this out for comments first. after discussion, I will do the following:

- add automated tests
- e2e test manually
- add the same changes in the `main` branch

### DIFF
--- a/finality-provider/cmd/fpd/daemon/commit_pr.go
+++ b/finality-provider/cmd/fpd/daemon/commit_pr.go
@@ -54,7 +54,6 @@ func runCommandCommitPubRand(ctx client.Context, cmd *cobra.Command, args []stri
 
 	// override to control the exact block height to commit to
 	cfg.MinRandHeightGap = 0
-	prCommitInterval := cfg.RandomnessCommitInterval
 
 	logger, err := log.NewRootLoggerWithFile(fpcfg.LogFile(homePath), cfg.LogLevel)
 	if err != nil {
@@ -76,7 +75,7 @@ func runCommandCommitPubRand(ctx client.Context, cmd *cobra.Command, args []stri
 		return fmt.Errorf("failed to get finality provider instance: %w", err)
 	}
 
-	return commitUntilHeight(fp, blkHeight, prCommitInterval)
+	return commitUntilHeight(fp, blkHeight, cfg.RandomnessCommitInterval)
 }
 
 func commitUntilHeight(
--- a/finality-provider/cmd/fpd/daemon/commit_pr.go
+++ b/finality-provider/cmd/fpd/daemon/commit_pr.go
@@ -1,0 +1,108 @@
+package daemon
+
+import (
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	bbntypes "github.com/babylonlabs-io/babylon/types"
+	fpcmd "github.com/babylonlabs-io/finality-provider/finality-provider/cmd"
+	fpcfg "github.com/babylonlabs-io/finality-provider/finality-provider/config"
+	"github.com/babylonlabs-io/finality-provider/finality-provider/service"
+	"github.com/babylonlabs-io/finality-provider/log"
+	"github.com/babylonlabs-io/finality-provider/util"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/spf13/cobra"
+)
+
+// CommandCommitPubRand returns the commit-pubrand command by connecting to the fpd daemon.
+func CommandCommitPubRand() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:     "commit-pubrand [fp-eots-pk-hex] [block-height]",
+		Aliases: []string{"cpr"},
+		Short:   "Manually trigger public randomness commitment for a finality provider",
+		Example: `fpd commit-pubrand --home /home/user/.fpd [fp-eots-pk-hex] [block-height]`,
+		Args:    cobra.ExactArgs(2),
+		RunE:    fpcmd.RunEWithClientCtx(runCommandCommitPubRand),
+	}
+	return cmd
+}
+
+func runCommandCommitPubRand(ctx client.Context, cmd *cobra.Command, args []string) error {
+	fpPk, err := bbntypes.NewBIP340PubKeyFromHex(args[0])
+	if err != nil {
+		return err
+	}
+	blkHeight, err := strconv.ParseUint(args[1], 10, 64)
+	if err != nil {
+		return err
+	}
+
+	// Get homePath from context like in start.go
+	clientCtx := client.GetClientContextFromCmd(cmd)
+	homePath, err := filepath.Abs(clientCtx.HomeDir)
+	if err != nil {
+		return err
+	}
+	homePath = util.CleanAndExpandPath(homePath)
+
+	cfg, err := fpcfg.LoadConfig(homePath)
+	if err != nil {
+		return fmt.Errorf("failed to load configuration: %w", err)
+	}
+
+	// override to control the exact block height to commit to
+	cfg.MinRandHeightGap = 0
+	prCommitInterval := cfg.RandomnessCommitInterval
+
+	logger, err := log.NewRootLoggerWithFile(fpcfg.LogFile(homePath), cfg.LogLevel)
+	if err != nil {
+		return fmt.Errorf("failed to initialize the logger: %w", err)
+	}
+
+	dbBackend, err := cfg.DatabaseConfig.GetDbBackend()
+	if err != nil {
+		return fmt.Errorf("failed to create db backend: %w", err)
+	}
+
+	fpApp, err := service.NewFinalityProviderAppFromConfig(cfg, dbBackend, logger)
+	if err != nil {
+		return fmt.Errorf("failed to create finality-provider app: %w", err)
+	}
+
+	fp, err := fpApp.GetFinalityProviderInstance(fpPk)
+	if err != nil {
+		return fmt.Errorf("failed to get finality provider instance: %w", err)
+	}
+
+	return commitUntilHeight(fp, blkHeight, prCommitInterval)
+}
+
+func commitUntilHeight(
+	fp *service.FinalityProviderInstance,
+	blkHeight uint64,
+	interval time.Duration,
+) error {
+	commitRandTicker := time.NewTicker(interval)
+	defer commitRandTicker.Stop()
+
+	lastCommittedHeight, err := fp.GetLastCommittedHeight()
+	if err != nil {
+		return fmt.Errorf("failed to get last committed height: %w", err)
+	}
+
+	for lastCommittedHeight < blkHeight {
+		<-commitRandTicker.C
+		_, err = fp.CommitPubRand(blkHeight)
+		if err != nil {
+			return fmt.Errorf("failed to commit public randomness: %w", err)
+		}
+
+		lastCommittedHeight, err = fp.GetLastCommittedHeight()
+		if err != nil {
+			return fmt.Errorf("failed to get last committed height: %w", err)
+		}
+	}
+	return nil
+}
--- a/finality-provider/service/fp_instance.go
+++ b/finality-provider/service/fp_instance.go
@@ -669,6 +669,7 @@ func (fp *FinalityProviderInstance) retryCommitPubRandUntilMaxRetry(targetBlockH
 // CommitPubRand generates a list of Schnorr rand pairs,
 // commits the public randomness for the managed finality providers,
 // and save the randomness pair to DB
+// Note: if the targetBlockHeight is too large, it will only commit fp.cfg.NumPubRand pairs
 func (fp *FinalityProviderInstance) CommitPubRand(targetBlockHeight uint64) (*types.TxResponse, error) {
 	lastCommittedHeight, err := fp.GetLastCommittedHeight()
 	if err != nil {
--- a/finality-provider/cmd/fpd/main.go
+++ b/finality-provider/cmd/fpd/main.go
@@ -34,6 +34,7 @@ func main() {
 		daemon.CommandGetDaemonInfo(), daemon.CommandCreateFP(), daemon.CommandLsFP(),
 		daemon.CommandInfoFP(), daemon.CommandRegisterFP(), daemon.CommandAddFinalitySig(),
 		daemon.CommandExportFP(), daemon.CommandTxs(),
+		daemon.CommandCommitPubRand(),
 	)
 
 	if err := cmd.Execute(); err != nil {


SebastianElvis:
Concept ACK, wdyt @gitferry @KonradStaniec

gitferry (commented):
I have mixed feelings about this. On one hand, I agree that removing fast sync when bootstraping is the right way to solve the lagging issue. On the other hand, adding this helper function is helpful in testing but also risky in the sense that public randomness is critical to security and users should not do it manually. @KonradStaniec wdyt?

gitferry:
> For security, what do you think are the additional risks if we add this? This command only commits more PRs. They still need to be timestamped properly, and there is no way to commit an incorrect PR pair since CommitPubRand was directly called

Hmmm tbh I can't think of any for now. Sounds good if it's for testing purpose only. Tbh we already have a riskier cmd [add-finality-sig](https://github.com/babylonlabs-io/finality-provider/blob/cb1c05ee7e349844916aeef521b3f61234b998b4/finality-provider/cmd/fpd/daemon/daemon_commands.go#L396) for testing. To make it more explicity, maybe we can have a switch in config to controll whether we enable such cmds or not, i.e., they are diabled by default

bap2pecs:
> To make it more explicity, maybe we can have a switch in config to controll whether we enable such cmds or not, i.e., they are diabled by default

sure but an easier way is to prefix with DANGEROUS. (i.e. not use suffix to avoid CLI auto-completion. so people has to type DANGEROUS first)

how about we name it to

```
fpd DANGEROUS_add-finality-sig
fpd DANGEROUS_commit-pub-rand
```

and add a WARNING in the description/help?

SebastianElvis:
What is the risk of manually committing public randomness? This won't lead to slashing or any other issue of the FP daemon, so does not incur much risk imo.

gitferry:
> sure but an easier way is to prefix with DANGEROUS. (i.e. not use suffix to avoid CLI auto-completion. so people has to type DANGEROUS first)

Sounds good to me

bap2pecs:
> What is the risk of manually committing public randomness? This won't lead to slashing or any other issue of the FP daemon, so does not incur much risk imo.

i guess the only risk is if you specify too high of the target PR commit height, you might drain the FP account BBN balance

KonradStaniec:
>I have mixed feelings about this. On one hand, I agree that removing fast sync when bootstraping is the right way to solve the lagging issue. On the other hand, adding this helper function is helpful in testing but also risky in the sense that public randomness is critical to security and users should not do it manually. @KonradStaniec wdyt?

Imo having commands for recovery is necessary, especially during early days of the prtocol where anything can happen.

I agree with with adding warning prefix to the commands which may be risky. Though maybe it should be `unsafe-..` for example `unsafe-add-finality-sig`. `DANGEROUS` is mouthful and have tricky spelling 😅  and `unsafe-` seems to be used in many contexts as warning.  We should also add docs for thaht i.e document why commands are unsafe and what are the risk of missuse.

bap2pecs:
> Though maybe it should be `unsafe-..` for example `unsafe-add-finality-sig`. `DANGEROUS` is mouthful and have tricky spelling 😅 and `unsafe-` seems to be used in many contexts as warning. We should also add docs for thaht i.e document why commands are unsafe and what are the risk of missuse.

agree. let's do `unsafe` and add docs. this should be good enough. anyone who start typing "unsafe-" should know what they r doing anyway lol

gitferry (commented):
I would recommend to have a separate method for manual submission like we did `TestSubmitFinalitySignatureAndExtractPrivKey` for manual submission of finality signatures. This way we have full flexibility of committing a range of randomness commit.

bap2pecs:
> I would recommend to have a separate method for manual submission like we did `TestSubmitFinalitySignatureAndExtractPrivKey` for manual submission of finality signatures. This way we have full flexibility of committing a range of randomness commit.

sg I can create a separate method. but it won't allow a range since we don't have any use case atm. the function will determine which height to start.

gitferry:
> sg I can create a separate method. but it won't allow a range since we don't have any use case atm. the function will determine which height to start.

how about we make args to optional flags, e.g., `--start-height` and `--rand-number`. If not set, the value will be retrieved from the consumer chain and config, respectively. Also, we should limit `rand-number` to be larger than `NumPubRand` in config. The cmd output should show the start height and end height (or rand number) of the commit

finality-provider/cmd/fpd/daemon/commit_pr.go
```diff
@@ -0,0 +1,79 @@
...
+
+// CommandCommitPubRand returns the commit-pubrand command by connecting to the fpd daemon.
+func CommandCommitPubRand() *cobra.Command {
+	var cmd = &cobra.Command{
+		// TODO: add --start-height as an optional flag
+		Use:     "unsafe-commit-pubrand [fp-eots-pk-hex] [block-height]",
+		Aliases: []string{"unsafe-cpr"},
+		Short:   "[UNSAFE] Manually trigger public randomness commitment for a finality provider",
+		Long: `[UNSAFE] Manually trigger public randomness commitment for a finality provider. ` +
+			`WARNING: this can drain the finality provider's balance if the block-height is too high.` +
```
gitferry, line 26:
```suggestion
		Long: `[UNSAFE] Manually trigger public randomness commitment for a finality provider.` +
			`WARNING: this can drain the finality provider's balance if the target-height is too high.` +
```

finality-provider/cmd/fpd/daemon/commit_pr.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/babylonlabs-io/finality-provider/log"
+	"github.com/babylonlabs-io/finality-provider/util"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/spf13/cobra"
+)
+
+// CommandCommitPubRand returns the commit-pubrand command by connecting to the fpd daemon.
+func CommandCommitPubRand() *cobra.Command {
+	var cmd = &cobra.Command{
+		// TODO: add --start-height as an optional flag
```
gitferry, line 21: I start to feel that adding `--start-height` as an option is important because of BTC-timestamping. A commit will be become effective after the committed epoch is finalized, which would take `100` BTC block time. Therefore, if the commit starts from the last committed height, which is behind, then all the randomness for the consumer blocks generated during `100` BTC block will be wasted

finality-provider/cmd/fpd/daemon/commit_pr.go
```diff
@@ -0,0 +1,79 @@
...
+		Use:     "unsafe-commit-pubrand [fp-eots-pk-hex] [block-height]",
+		Aliases: []string{"unsafe-cpr"},
+		Short:   "[UNSAFE] Manually trigger public randomness commitment for a finality provider",
+		Long: `[UNSAFE] Manually trigger public randomness commitment for a finality provider. ` +
+			`WARNING: this can drain the finality provider's balance if the block-height is too high.` +
+			`Note: if there is no pubrand committed before, it will only commit the pubrand for the target block-height.`,
```
gitferry, line 27: In this case, we probably need the user to specify `--start-height` as this is probably a mistake. Making a commit for a single height does not make much sense unless the user explicity specifies it

finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -327,10 +327,11 @@ func runCommandRegisterFP(cmd *cobra.Command, args []string) error {
...
 	var cmd = &cobra.Command{
-		Use:     "add-finality-sig [fp-eots-pk-hex] [block-height]",
-		Aliases: []string{"afs"},
-		Short:   "Send a finality signature to the consumer chain. This command should only be used for presentation/testing purposes",
-		Example: fmt.Sprintf(`fpd add-finality-sig --daemon-address %s`, defaultFpdDaemonAddress),
+		Use:     "unsafe-add-finality-sig [fp-eots-pk-hex] [block-height]",
+		Aliases: []string{"unsafe-afs"},
+		Short:   "[UNSAFE] Send a finality signature to the consumer chain.",
+		Long:    "[UNSAFE] Send a finality signature to the consumer chain. This command should only be used for presentation/testing purposes",
+		Example: fmt.Sprintf(`fpd unsafe-add-finality-sig --daemon-address %s`, defaultFpdDaemonAddress),
```
gitferry, line 334:
```suggestion
		Example: fmt.Sprintf(`fpd unsafe-add-finality-sig [fp-eots-pk-hex] [block-height] --daemon-address %s`, defaultFpdDaemonAddress),
```

finality-provider/service/fp_instance.go
```diff
@@ -724,12 +733,60 @@ func (fp *FinalityProviderInstance) CommitPubRand(targetBlockHeight uint64) (*ty
 
 	// Update metrics
 	fp.metrics.RecordFpRandomnessTime(fp.GetBtcPkHex())
-	fp.metrics.RecordFpLastCommittedRandomnessHeight(fp.GetBtcPkHex(), lastCommittedHeight)
```
gitferry, line 727: we'd better also record the local last committed height to the metrics

finality-provider/service/fp_instance.go
```diff
@@ -724,12 +733,60 @@ func (fp *FinalityProviderInstance) CommitPubRand(targetBlockHeight uint64) (*ty
...
 	fp.metrics.AddToFpTotalCommittedRandomness(fp.GetBtcPkHex(), float64(len(pubRandList)))
 
 	return res, nil
 }
 
+// TestCommitPubRand is exposed for devops/testing purpose to allow manual committing public randomness in cases
+// where FP is stuck due to lack of public randomness (https://github.com/babylonlabs-io/finality-provider/issues/115).
```
gitferry, line 742: seems the issue is no longer valid?

finality-provider/service/fp_instance.go
```diff
@@ -724,12 +733,60 @@ func (fp *FinalityProviderInstance) CommitPubRand(targetBlockHeight uint64) (*ty
...
+// Note:
+// - this function is similar to `CommitPubRand` but should not be used in the main pubrand submission loop.
+// - it will always start from the last committed height + 1
+// - if targetBlockHeight is too large, it will commit multiple fp.cfg.NumPubRand pairs in a loop until reaching the targetBlockHeight
+func (fp *FinalityProviderInstance) TestCommitPubRand(targetBlockHeight uint64) error {
+	lastCommittedHeight, err := fp.GetLastCommittedHeight()
+	if err != nil {
+		return err
+	}
+	fp.metrics.RecordFpLastCommittedRandomnessHeight(fp.GetBtcPkHex(), lastCommittedHeight)
```
gitferry, line 753: seems more accurate to record the local last committed height after the commit is made

finality-provider/service/fp_instance.go
```diff
@@ -724,12 +733,60 @@ func (fp *FinalityProviderInstance) CommitPubRand(targetBlockHeight uint64) (*ty
...
+			"finality provider has already committed pubrand to target block height (pk: %s, target: %d, last committed: %d)",
+			fp.GetBtcPkHex(),
+			targetBlockHeight,
+			lastCommittedHeight,
+		)
+	}
+	fmt.Printf("Start committing pubrand from block height %d...\n", startHeight)
+
+	commitRandTicker := time.NewTicker(fp.cfg.RandomnessCommitInterval)
+	defer commitRandTicker.Stop()
```
gitferry, line 773: I was thinking we should not let user to make the commit with number of randomness larger than the value in config. This way, we can either reject large `target-height` to avoid draining user's funds

bap2pecs: this command won't be used if it can only commit e.g. 1000 PRs at once. I think the risk is manageable given there is already unsafe prefix and documentation to explain the risk. the idea is anyone who use the command should know exactly what they are doing

finality-provider/cmd/fpd/daemon/commit_pr.go
bap2pecs: good catch. this docstring needs to be fixed. it's no longer the case with the new updates in this PR

finality-provider/service/fp_instance.go
bap2pecs: yes it's still recorded. just moved due to refactoring

finality-provider/service/fp_instance.go
gitferry: Agreed but why do we need to set interval? Can the commits be bundled in a single tx with several commit messages?

bap2pecs: there is already `NumPubRand` in the config. each commit tx will commit this many PRs. the user can change the size in the fpd.conf file.

gitferry: For example, the user will commit `10k` randomness while the config `NumPubRand=1000`, so the cmd will send 10 tx and the user needs to wait for at least `10 * fp.cfg.RandomnessCommitInterval` before it finishes. Why do we want the user to wait for `fp.cfg.RnadomnessCommitInterval` for each commit?

bap2pecs: makes sense. in this case, the user can set the interval to 0 or sth very small. if I am gonna run it myself, I will probably set it to at least 5 seconds just to be safe

bap2pecs: maybe it's not necessary but i was thinking about some potential nonce mismatch issue. if we are talking about 2sec block time on the consumer chain, there will be 3600*24/2=43200 blocks per day.

if I set `NumPubRand` to 10000, it should only take 20sec even I add this wait time. and committing PRs for one month's usage would only take 10min. so I guess it's tolerable anyway

gitferry: why would there be nonce mismatch issue? I thought this issue will occur only when we send txs concurrently

bap2pecs: iiuc, FP interacts with the keyring backend to send txs, if FP sends txs too fast, I am not sure whether there will be any race conditions. maybe the Cosmos keyring infra is quite robust but I am not too familiar w it

gitferry: FP relies on Babylon client implementation which has access control over the keyring `https://github.com/babylonlabs-io/babylon/blob/89dd901bb40dc2420ad196591ec6babe6339cd3c/client/client/tx.go#L50`, so I think it's safe to remove the sending interval. Maybe a better way is to bundle all the commit messages into one like we do for batch finality signatures but I guess we can do it for future optimization

bap2pecs: got it. I will add a TODO

bap2pecs: i tested and it doesn't work after removing the interval. i do meet with the nonce mismatch err

```
2024-11-21T01:26:54.093326Z	info	Start committing pubrand from block height	{"start_height": 900}
2024-11-21T01:27:02.632290Z	debug	retrying	{"attemp": 1, "max_attempts": 5, "error": "rpc error: code = Unknown desc = rpc error: code = Unknown desc = account sequence mismatch, expected 11, got 10: incorrect account sequence [cosmos/cosmos-sdk@v0.50.9/x/auth/ante/sigverify.go:290] with gas used: '19626': unknown request"}
2024-11-21T01:27:10.649841Z	debug	retrying	{"attemp": 2, "max_attempts": 5, "error": "rpc error: code = Unknown desc = rpc error: code = Unknown desc = failed to execute message; message index: 0: The start height (900) has overlap with the height of the highest public randomness committed (999): execute wasm contract failed [CosmWasm/wasmd@v0.51.0/x/wasm/keeper/keeper.go:422] with gas used: '114308': unknown request"}
2024-11-21T01:27:19.133247Z	debug	retrying	{"attemp": 3, "max_attempts": 5, "error": "rpc error: code = Unknown desc = rpc error: code = Unknown desc = failed to execute message; message index: 0: The start height (900) has overlap with the height of the highest public randomness committed (999): execute wasm contract failed [CosmWasm/wasmd@v0.51.0/x/wasm/keeper/keeper.go:422] with gas used: '114308': unknown request"}
```

gitferry: It seems the error is because the fpd is already running and committing randomness while you executed the command. This will cause concurrent submission

bap2pecs: > It seems the error is because the fpd is already running and committing randomness while you executed the command. This will cause concurrent submission

I don't think so. this new command doesn't require `fpd daemon` running

gitferry: I see. But why the account mismatch error occurs at the first submission?

bap2pecs: I don't know exactly why but i think this is what happened:

- started PR commit at height 900
- first tx is sent for PR 900-999
- before it's processed by the Babylon chain, the second tx for PR 1000-1099 is sent
- but since the first tx is not finished, it reuses the same nonce so we see this error before we see the "Successful transaction" log

gitferry: I see. This is indeed weried as the txs should be sent sequentially. Anyway, let's leave the interval then. Maybe we can also leave a note or TODO there?

gitferry: btw the unit tests failure might be fixed by https://github.com/babylonlabs-io/finality-provider/pull/142

bap2pecs:
@gitferry alright, i addressed all your comments and tested the command. see the updated "Test Plan" section for the test setup

finality-provider/service/fp_instance.go
bap2pecs: approved. i will test again after #142 is merged

gitferry: Thanks. Merged

bap2pecs: tested. nonce issue is fixed now. thanks!

finality-provider/cmd/fpd/daemon/commit_pr.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/babylonlabs-io/finality-provider/metrics"
+	"github.com/babylonlabs-io/finality-provider/util"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/spf13/cobra"
+)
+
+// CommandCommitPubRand returns the commit-pubrand command by connecting to the fpd daemon.
+func CommandCommitPubRand() *cobra.Command {
+	var cmd = &cobra.Command{
+		// TODO: add --start-height as an optional flag
```
gitferry, line 26: this can be removed

finality-provider/service/fp_instance.go
```diff
@@ -669,6 +709,9 @@ func (fp *FinalityProviderInstance) retryCommitPubRandUntilMaxRetry(targetBlockH
 // CommitPubRand generates a list of Schnorr rand pairs,
 // commits the public randomness for the managed finality providers,
 // and save the randomness pair to DB
+// Note:
+// - if there is no pubrand committed before, it will start from the targetBlockHeight
```
gitferry, line 713: in the code, it seems it starts from `0` but either way it should start from a future height considering btc timestamping latency. See https://github.com/babylonlabs-io/finality-provider/blob/main/docs/commit-pub-rand.md#determining-start-height

gitferry (approved):
Awesome work! Thanks for all the discussions and changes. Do you also want to back-port it to `main`?

bap2pecs:
> Awesome work! Thanks for all the discussions and changes. Do you also want to back-port it to `main`?

absolutely. i have that in the RP description as the Next Step :)

thanks for all the discussions and feedback!

PR #132, opened 2024-11-19
chore: Replace fast sync with batch processing

### LINKED ISSUE #118: Cherry-pick batch submission to replace fast sync
* Cherry-pick the batch submission change https://github.com/babylonlabs-io/finality-provider/pull/57 or do it from scratch
* Remove fast sync completely

gitferry:
Closes #118. Notable changes:
* remove fast sync completely
* introduce batch process for catching up
* introduce signature submission interval to config
* introduce batch submission size to config


finality-provider/service/fp_instance.go
```diff
@@ -814,36 +683,46 @@ func (fp *FinalityProviderInstance) TestSubmitFinalitySignatureAndExtractPrivKey
...
-		if latestFinalisedBlock[0].Height > initialBlockToGet {
-			initialBlockToGet = latestFinalisedBlock[0].Height
-		}
+
+	// start from finality activation height
+	startHeight := finalityActivationHeight
+
+	latestFinalisedBlocks, err := fp.latestFinalizedBlocksWithRetry(1)
+	if err != nil {
+		return 0, err
```
Lazar955, line 713: nit: wrap errors

Lazar955 (approved):
lgtm! Good job simplifying the logic!

finality-provider/config/config.go
```diff
@@ -18,26 +18,24 @@ import (
...
+	defaultNumPubRandMax               = 100000
+	defaultMinRandHeightGap            = 35000
+	defaultStatusUpdateInterval        = 20 * time.Second
+	defaultRandomInterval              = 30 * time.Second
+	defaultSubmitRetryInterval         = 1 * time.Second
+	defaultSyncFpStatusInterval        = 30 * time.Second
+	defaultSignatureSubmissionInterval = 1 * time.Second
+	defaultMaxSubmissionRetries        = 20
+	defaultBitcoinNetwork              = "signet"
+	defaultDataDirname                 = "data"
```
bap2pecs, line 38: Shouldn't we add a config for batch processing size? what's the current code behavior if it lags way behind?

finality-provider/config/poller.go
```diff
@@ -4,7 +4,7 @@ import "time"
 
 var (
 	defaultBufferSize        = uint32(1000)
-	defaultPollingInterval   = 20 * time.Second
+	defaultPollingInterval   = 100 * time.Millisecond
```
bap2pecs, line 7: i think the polling interval shouldn't be less than the chain block-producing time. OP chain is usually 2-3 seconds. what's the block time for Babylon chain?

finality-provider/service/fp_instance.go
```diff
@@ -246,53 +186,64 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
...
+		}
+	}
+}
+
+func (fp *FinalityProviderInstance) getAllBlocksFromChan() []*types.BlockInfo {
+	var pollerBlocks []*types.BlockInfo
+	for {
+		select {
+		case b := <-fp.poller.GetBlockInfoChan():
+			shouldProcess, err := fp.shouldProcessBlock(b)
```
bap2pecs, line 209: there might be a problem here since `shouldProcessBlock` makes RPC calls to the Babylon chain. so in this `for` loop we are doing a lot of RPC calls.

finality-provider/service/fp_instance.go
```diff
@@ -453,29 +319,32 @@ func (fp *FinalityProviderInstance) reportCriticalErr(err error) {
...
 	// error will be returned if maximum retries have been reached or the query to the consumer chain fails
 	for {
 		select {
 		case <-time.After(fp.cfg.SubmissionRetryInterval):
 			// error will be returned if max retries have been reached
-			res, err := fp.SubmitFinalitySignature(targetBlock)
+			var res *types.TxResponse
+			var err error
+			if len(targetBlocks) == 1 {
+				res, err = fp.SubmitFinalitySignature(targetBlocks[0])
```
bap2pecs, line 337: we can probably get rid of `SubmitFinalitySignature` from the interface

finality-provider/service/fp_instance.go
```diff
@@ -912,6 +791,28 @@ func (fp *FinalityProviderInstance) latestFinalizedBlocksWithRetry(count uint64)
 	return response, nil
 }
 
+func (fp *FinalityProviderInstance) finalityActivationHeightWithRetry() (uint64, error) {
```
bap2pecs, line 794: nit: rename to `getFinalityActivationHeightWithRetry`

finality-provider/service/fp_instance.go
gitferry: I would keep the interface but the implementation could just use the one from the batch submission

finality-provider/service/fp_instance.go
gitferry: Agreed this case could happen in case of catching up. Left a TODO here for now

finality-provider/config/poller.go
gitferry: I would argue that we should have a short polling interval for a quick catching up. Would `ms` level interval a concern in practice?

finality-provider/config/config.go
gitferry: Good point. Fixed.

gitferry: If the fp lags behind, the poller should quickly poll blocks (that's why we should have a short polling interval). Then the fp will do batch submission

gitferry:
I would wait for one more approval before merge

finality-provider/config/poller.go
bap2pecs: Catching up is not a normal case. If it's way behind, people should just change the config file temporarily to speed up. imo default value should be for normal  submission cases

bap2pecs (approved):
lgtm. can we cherrypick this to the base consumer branch?

finality-provider/config/poller.go
gitferry: Makes sense. Changed it to `1s` and added notes there

gitferry:
> lgtm. can we cherrypick this to the base consumer branch?

Sure.

finality-provider/service/fp_instance.go
```diff
@@ -246,53 +186,68 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
...
+			fp.logger.Info("the finality signature submission loop is closing")
+			return
+		}
+	}
+}
+
+func (fp *FinalityProviderInstance) getAllBlocksFromChan() []*types.BlockInfo {
+	var pollerBlocks []*types.BlockInfo
+	for {
+		select {
```
KonradStaniec, line 207: I think you could also add:
```
case <-fp.quit:
```

to this select, to not wait long time for quit in case of many blocks pending on poller channel.

KonradStaniec (approved):
looks much cleaner now  👍

bap2pecs:
@gitferry do you plan to backport this to the base branch?

gitferry:
@bap2pecs not sure it's efficient to backport every change to the base branch. Do we have a routine to periodically backport changes to the base branch?

bap2pecs:
> @bap2pecs not sure it's efficient to backport every change to the base branch. Do we have a routine to periodically backport changes to the base branch?

actually it's probably the opposite. I feel it's probably always better to backport right away or context will be lost. for example, https://github.com/babylonlabs-io/finality-provider/pull/153 is not a straightforward backport due to some interface level difference between main and base branch. but I can quickly backport it b/c my memory was fresh so I know where to change

also, i'd say for important changes like this, we should proritize. this one is currently blocking some of our works (e.g. e2e test refactoring) so we'd appreciate if it can be backported

bap2pecs:
> @bap2pecs Make sense. I'll backport it

thanks a lot! I remember @vitsalis   told me in Bangkok that this backport issue will be solved soon but i don't remember the conversation

do you guys feel that backporting takes lots of engineering time for your team? @gitferry @SebastianElvis

PR #145, opened 2024-11-21
fix: Voting power query error handler for consumer chain

gitferry:
The previous fix #142 only fixed the error handler for babylon controller chain other than the cosumer chain. This PR addressed it.


bap2pecs (approved):

gitferry:
@bap2pecs Do you have any ideas why e2e_op failed due to this change?

SebastianElvis (approved):

gitferry:
Changed `cfg.NumPubRand` from 64 to 100 and it worked

PR #148, opened 2024-11-22
fix: allow to create keys with empty hd path

RafilxTenfen:
- Copy the `keys add` command from cosmos-sdk just modified to allow empty HD path (The choice to continue to use the SDK is due the rest of the commands import/export work as expected)
- Move cmd files from package main to eotsd, it helps to debbug


eotsmanager/cmd/eotsd/daemon/keysadd.go
```diff
@@ -0,0 +1,396 @@
...
+	"github.com/cosmos/cosmos-sdk/crypto/hd"
+	cryptokeyring "github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/cosmos/cosmos-sdk/crypto/keys/multisig"
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/go-bip39"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+)
+
```
gitferry, line 29: maybe we can highlight somewhere that this part of code is adapted from Cosmos SDK and we add a perminent link here?

eotsmanager/cmd/eotsd/daemon/keysadd.go
```diff
@@ -0,0 +1,396 @@
...
+			if err != nil {
+				return err
+			}
+
+			if bip39Passphrase != p2 {
+				return errors.New("passphrases don't match")
+			}
+		}
+	}
+
```
gitferry, line 278: So it is where we made the modification, right (removing the hdPath empty check)? Let's highlight what we modified here for future reference

gitferry (approved):
Great work!

eotsmanager/cmd/eotsd/daemon/keysadd.go
RafilxTenfen: done at https://github.com/babylonlabs-io/finality-provider/pull/148/commits/e03230b89f5ab74fc9eb189df0e9385e9ca41e1f

eotsmanager/cmd/eotsd/daemon/keysadd.go
RafilxTenfen: done at https://github.com/babylonlabs-io/finality-provider/pull/148/commits/e03230b89f5ab74fc9eb189df0e9385e9ca41e1f

PR #149, opened 2024-11-22
chore: remove update of config after keys add in fpd

RafilxTenfen:


bap2pecs (approved):
thanks for the fix! can you also cherry-pick this to the base consumer branch?

gitferry (approved):

Lazar955 (approved):

PR #150, opened 2024-11-22
chore: `QueryFinalityProviderHasPower` returns `false` if there is no public randomness

lesterli:
## Summary

This PR updates the function `QueryFinalityProviderHasPower`, it will return `false` if there is no public randomness at the queried height.

More content: https://github.com/babylonlabs-io/finality-provider/issues/115#issuecomment-2484624315

## Test Plan

open the debug log in `op_test_manager.go`, line 74: zapcore.InfoLevel -> zapcore.DebugLevel, and then run e2e test:
```
make test-e2e-op-filter FILTER=TestOpFpNoVotingPower
```
once it is done, it should have some debug logs, like this but the value of fp pk is different:
```
    utils.go:100: Public randomness for fp 9ff0d9db0a73b7d4b3aa19b0b6c7f56ec4089ea22194a6b5019f021b8d342487 is successfully committed at height 64
2024-11-22T11:08:30.761+0800    DEBUG   opstackl2/consumer.go:285       FP last committed public randomness     {"height": 64}
2024-11-22T11:08:30.761+0800    DEBUG   opstackl2/consumer.go:290       FP has 0 voting power because there is no public randomness at this height      {"height": 65}

2024-11-22T11:08:31.772+0800    DEBUG   opstackl2/consumer.go:285       FP last committed public randomness     {"height": 64}
2024-11-22T11:08:31.774+0800    DEBUG   opstackl2/consumer.go:326       FP has 0 voting power because there is no BTC delegation        {"fp_btc_pk": "9ff0d9db0a73b7d4b3aa19b0b6c7f56ec4089ea22194a6b5019f021b8d342487", "height": 1}
2024-11-22T11:08:31.774+0800    DEBUG   service/fp_instance.go:474      the finality-provider does not have voting power        {"pk": "9ff0d9db0a73b7d4b3aa19b0b6c7f56ec4089ea22194a6b5019f021b8d342487", "block_height": 1}
```

Note: For the op e2e test, there is a working branch `test/fix-fg-circular-dependency-e2e-tests`


clientcontroller/opstackl2/consumer.go
```diff
@@ -264,13 +271,31 @@ func (cc *OPStackL2ConsumerController) SubmitBatchFinalitySigs(
...
+	pubRand, err := cc.QueryLastPublicRandCommit(fpPk)
+	if err != nil {
+		return false, err
+	}
+	if pubRand == nil {
+		return false, nil
+	}
+
+	// fp has 0 voting power if there is no public randomness at this height
+	lastCommittedPubRandHeight := pubRand.StartHeight + pubRand.NumPubRand - 1
```
SebastianElvis, line 284: Nit: `pubRand.StartHeight + pubRand.NumPubRand - 1` can be a function `EndHeight()` under `pubRand`

SebastianElvis (approved):

gitferry (approved):

clientcontroller/opstackl2/consumer.go
```diff
@@ -264,13 +271,31 @@ func (cc *OPStackL2ConsumerController) SubmitBatchFinalitySigs(
...
+		return false, nil
+	}
+
+	// fp has 0 voting power if there is no public randomness at this height
+	lastCommittedPubRandHeight := pubRand.EndHeight()
+	cc.logger.Debug(
+		"FP last committed public randomness",
+		zap.Uint64("height", lastCommittedPubRandHeight),
+	)
+	if blockHeight > lastCommittedPubRandHeight {
```
bap2pecs, line 289: this is good but there is another case that this check won't cover

what if pubrand is not consecutive? so block 1-100 and 300-400 has PRs. if `blockHeight` is 200, `lastCommittedPubRandHeight` will be 400 and it should be considered not having VP

but this check doesn't cover it

go.mod
```diff
@@ -446,9 +446,12 @@ replace (
...
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+	// fix for `make build` error: invalid CFI advance_loc expression
+	github.com/supranational/blst => github.com/supranational/blst v0.3.13
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
-
 	// avoid v1.66 that has a breaking change for protobuf. That change breaks the relayer.
 	// https://github.com/grpc/grpc-go/issues/7569
 	google.golang.org/grpc => google.golang.org/grpc v1.65.0
+	// fix for `go mod tidy` error: unknown revision v1.8.6
+	nhooyr.io/websocket => github.com/coder/websocket v1.8.7
```
bap2pecs, line 456: just curious, why previous CI run never detect these?

itest/opstackl2/op_e2e_test.go
```diff
@@ -26,8 +26,20 @@ func TestOpFpNoVotingPower(t *testing.T) {
...
-	// query the first committed pub rand
 	opcc := ctm.getOpCCAtIndex(0)
+
+	// query the last committed pub rand
+	lastCommittedPubRand, err := queryFirstOrLastPublicRandCommit(opcc, fpInstance.GetBtcPk(), false)
+	require.NoError(t, err)
+	require.NotNil(t, lastCommittedPubRand)
+
+	// no public randomness at this height, so the FP has no voting power
+	nextLastCommittedPubRandHeight := lastCommittedPubRand.EndHeight() + 1
```
bap2pecs, line 37: nextLastCommittedPubRandHeight -> blockHeightWithNoPubRand

bap2pecs (approved):

clientcontroller/opstackl2/consumer.go
gitferry: Good point as Babylon allows gaps. Better to add a TODO here.

Also, currently BTC-timestamping is not involved, right?

bap2pecs: yes it doesn't need to be in this PR. we can add a TODO

no currently BTC ts is not involved

go.mod
lesterli: yes, i tested on a new server with ubuntu 22.04, and it works fine.

for `go mod tidy`, i saw it automatically selects the appropriate version on this new server
```
go: downloading nhooyr.io/websocket v1.8.17
```
for `make build`, the error only occurs on x86 macOS using LLVM 17

so i will revert these changes.

PR #167, opened 2024-11-27
chore: Remove last processed height

gitferry:
The `lastProcessedHeight` was introduced for fast catch up blocks for which the fp does not have voting power. However, this case rarely happens and is unharmful. More discussions can be found [here](https://github.com/babylonlabs-io/finality-provider/pull/120#discussion_r1836699698).


finality-provider/service/fp_instance.go
```diff
@@ -685,9 +675,8 @@ func (fp *FinalityProviderInstance) getPollerStartingHeight() (uint64, error) {
 		startHeight = fppath.MaxUint64(startHeight, latestFinalisedBlocks[0].Height+1)
 	}
 
-	// consider the height after the last processed block
-	lastProcessedHeight := fp.GetLastProcessedHeight()
-	startHeight = fppath.MaxUint64(startHeight, lastProcessedHeight+1)
+	// consider the height after the last voted height
+	startHeight = fppath.MaxUint64(startHeight, fp.GetLastVotedHeight()+1)
```
Lazar955, line 679:
```suggestion
	startHeight = max[uint64](startHeight, fp.GetLastVotedHeight()+1)
```

Lazar955: Nit, let's remove this util function and use go's built-in one. I see that we always pass 2 arguments anyways

Lazar955 (approved):
Nice cleanup!

finality-provider/service/fp_instance.go
gitferry: Good point!

PR #182, opened 2024-12-03
chore: Remove fp manager

### LINKED ISSUE #112: Remove finality provider manager
It is not needed after the removal of support of running multiple finality provider instances

gitferry:
Closes #112. This PR also did some cleanup:
- move all the loops to `event_loops.go`
- remove legacy code


finality-provider/service/errors.go
```diff
@@ -1,6 +1,22 @@
...
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+
+	bbntypes "github.com/babylonlabs-io/babylon/types"
+)
+
+const instanceTerminatingMsg = "terminating the finality-provider instance due to critical error"
```
SebastianElvis, line 10: Why is this not a type of `CriticalError`?

SebastianElvis (approved):
Much cleaner!

finality-provider/service/errors.go
gitferry: Did you mean why the message not part of `CriticalError`? This is because this is only used in https://github.com/babylonlabs-io/finality-provider/blob/64ae9cacff4d91919391cc99c6c630d3649f06a9/finality-provider/service/event_loops.go#L177 where we decide to terminate the instance. Not for all of the cases

PR #193, opened 2024-12-04
chore: adds unsafeSignEOTS for testing double signing

Lazar955:


eotsmanager/eotsmanager.go
```diff
@@ -31,6 +31,9 @@ type EOTSManager interface {
 	// or passPhrase is incorrect
 	SignEOTS(uid []byte, chainID []byte, msg []byte, height uint64, passphrase string) (*btcec.ModNScalar, error)
 
+	// UnsafeSignEOTS should only be used in e2e tests for demonstration purposes. Use SignEOTS for real operations
```
gitferry, line 34: We should mention that `UnsafeSignEOTS` does not check double-sign

eotsmanager/eotsmanager.go
```diff
@@ -31,6 +31,9 @@ type EOTSManager interface {
 	// or passPhrase is incorrect
 	SignEOTS(uid []byte, chainID []byte, msg []byte, height uint64, passphrase string) (*btcec.ModNScalar, error)
```
gitferry, line 32: We can mention that `SignEOTS` should have built-in anti-slashing mechanism to ensure signature for the same height will not be signed twice

gitferry (approved):
Great work!

itest/e2e_test.go
```diff
@@ -98,14 +98,11 @@ func TestDoubleSigning(t *testing.T) {
...
 	}
-	_, _, err = fpIns.TestSubmitFinalitySignatureAndExtractPrivKey(b)
-	require.Error(t, err)
-
-	// todo(lazar): implement UnsafeSignEOTS, containing logic without double sign protection
-	//require.NoError(t, err)
-	//require.NotNil(t, extractedKey)
-	//localKey := tm.GetFpPrivKey(t, fpIns.GetBtcPkBIP340().MustMarshal())
-	//require.True(t, localKey.Key.Equals(&extractedKey.Key) || localKey.Key.Negate().Equals(&extractedKey.Key))
+	_, extractedKey, err = fpIns.TestSubmitFinalitySignatureAndExtractPrivKey(b)
```
gitferry, line 101: I was thinking that maybe we should keep an option to use `SignEOTS` within `TestSubmitFinalitySignatureAndExtractPrivKey` other than only the unsafe one so that we can test with the slashing protection works

gitferry (commented):
Could you also add logs when duplicated vote from db is sent?

PR #199, opened 2024-12-06
fix: EOTS signing for multiple finality providers

gitferry:
To support EOTS signing for multiple finality providers, the signing record should be saved with key (public key || chainID || height)


eotsmanager/store/storedsigningrecord.go
```diff
@@ -1,17 +1,29 @@
...
-	s.BlockHash = sr.BlockHash
+	s.Msg = sr.Msg
 	s.Timestamp = sr.Timestamp
-	s.Signature = sr.Signature
+	s.Signature = sr.EotsSig
+}
+
+func getSignRecordKey(pk, chainID []byte, height uint64) []byte {
+	// calculate the randomn hash of the key concatenated with chainID and height
+	digest := hmac.New(sha256.New, pk)
```
KonradStaniec, line 26: hmm I wonder whether just concatenating bytes wouldn't be be better as key.

If the key is `fpPk || height || chainID` then finding all records for `fp` is just finding all keys with prefix `fpPk || ..`. I am not sure what this hashing gives us except decreasing key slightly. 

(this is not blocking though, we can disscuss it in separate issue)

KonradStaniec (approved):

eotsmanager/store/storedsigningrecord.go
gitferry: I agree. Will do it in this PR as db schema change in the future is troublesome

gitferry: Done in [48ac5ee](https://github.com/babylonlabs-io/finality-provider/pull/199/commits/48ac5eea80c7c9ff9f994efa830c09aa1059b999)

PR #207, opened 2024-12-09
chore: create fp from json

Lazar955:
Alternative to passing in flags, the operator would provide a JSON file.

[Closes](https://github.com/babylonlabs-io/finality-provider/issues/14)

### DIFF
--- a/finality-provider/cmd/fpd/daemon/daemon_commands.go
+++ b/finality-provider/cmd/fpd/daemon/daemon_commands.go
@@ -167,6 +167,16 @@ func runCommandCreateFP(ctx client.Context, cmd *cobra.Command, _ []string) erro
 		}
 	}
 
+	// read flag regardless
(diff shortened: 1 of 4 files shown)


finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -516,3 +517,155 @@ func loadKeyName(homeDir string, cmd *cobra.Command) (string, error) {
 	}
 	return keyName, nil
 }
+
+type parsedFinalityProvider struct {
+	keyName        string
+	chainID        string
+	eotsPK         string
+	passphrase     string
+	daemonAddress  string
```
gitferry, line 526: Not sure this should be part of it. Can we make it a flag either way?

gitferry (approved):
Great work! A minor comment. @samricotta Let's modify the instructions accordingly

PR #208, opened 2024-12-10
chore: Remove sync fp status loop

### LINKED ISSUE #191: Remove `SyncFinalityProviderStatus`
This is legacy when fps could be injected to Babylon during upgrade. Keeping it is bug-prune

gitferry:
Closes #191. In particular, this PR:
- removed the loop for syncing fp status but syncing in the main thread during fp app initiation.
  - the loop is no longer needed as we have decided no fp will be registered via chain upgrade
  - we still need syncing status during initiation in case the fp is slashed or jailed
- in `monitorStatusUpdate`, the check for jailing and slashing is removed. This is because the removal of voting power due to slashing or jailing happens in the next block. Countering jailing and slashing in the loop will cause the fp to not vote when it still have voting power for the current block

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+### Improvements
+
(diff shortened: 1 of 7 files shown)


Lazar955 (approved):
lgtm, nice work!

RafilxTenfen (approved):
LGTM :tada:

PR #211, opened 2024-12-10
chore: Clean up unused cmd

### LINKED ISSUE #163: Remove all the cmd that will not be used for Phase-2

gitferry:
Closes #163. In particular:

- removed `fpd tx` command set but moved `NewWithdrawRewardCmd` and `NewSetWithdrawAddressCmd` to the root
- removed default value of `--commission-rate` as it is a required flag
- removed rpc `SignMessageFromChainKey` and relevant functionalities
- removed `fpd export-finality-provider`


finality-provider/cmd/fpd/daemon/export.go
RafilxTenfen, line 1: I would leave the export command

RafilxTenfen (approved):
utACK

finality-provider/cmd/fpd/daemon/export.go
gitferry: what would be the use case?

RafilxTenfen: Just to see the information based on the finality provider eots-pk, the signing part could be removed

gitferry: we have `fpd finality-provider-info [fp-eots-pk-hex]` for that

RafilxTenfen: true

PR #155, opened 2024-11-25
backport: Replace fast sync with batch processing

gitferry:
Notable changes:
* remove fast sync completely
* introduce batch process for catching up
* introduce signature submission interval to config
* introduce batch submission size to config

PR #213, opened 2024-12-11
fix: ChainType choices (to support consumer chains)

bap2pecs:
## Summary

fix a small bug which was introduced in https://github.com/babylonlabs-io/finality-provider/pull/155

this caused FP failed to start

```
bash-5.1$ fpd start
service injective.evm.v1beta1.Msg does not have cosmos.msg.v1.service proto annotation
Error: failed to load configuration: /home/finality-provider/.fpd/fpd.conf:6: Invalid value `OPStackL2' for option `--chaintype'. Allowed values are: 
Usage:
  fpd start [flags]
```

## Test Plan

tested with a new docker image created by CI
```
babylonlabs/finality-provider:9fcbde4fdacd26de256a944dedacb645826c2e9e
```

and it doesn't have this error anymore

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,58 +6,58 @@ on:
       - '**'
 
 jobs:
-  lint_test:
-    uses: babylonlabs-io/.github/.github/workflows/reusable_go_lint_test.yml@v0.7.0
-    secrets: inherit
-    with:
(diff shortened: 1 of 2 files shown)


SebastianElvis (approved):

bap2pecs:
image

this one still seems flaky. rerunning.

PR #215, opened 2024-12-11
docs: fix invalid link README.md

osrm:
# Rationale for this change

There is invalid link in the document.

# Changes

Fix invalid link in the document. The commit explains which link is invalid and how it is fixed.


gitferry (approved):
Thanks for the fix!

PR #217, opened 2024-12-12
fix: shouldn't wait `SubmissionRetryInterval` at the first iteration

bap2pecs:
## Summary

we found an issue in our OP devnet that finality sigs were submitted very slow. the root cause is we shouldn't wait 300 seconds in the first iteration in the for loop inside `retrySubmitSigsUntilFinalized()`

more context: https://github.com/babylonlabs-io/finality-provider/pull/155#discussion_r1881988164

## Test Plan

tested in our devnet with the new image

set `PollInterval` to 0.5 sec and see how it "fast synced" to the latest block

image

### DIFF
--- a/finality-provider/service/fp_instance.go
+++ b/finality-provider/service/fp_instance.go
@@ -403,58 +403,55 @@ func (fp *FinalityProviderInstance) retrySubmitSigsUntilFinalized(targetBlocks [
 	var failedCycles uint32
 	targetHeight := targetBlocks[len(targetBlocks)-1].Height
 
-	// we break the for loop if the block is finalized or the signature is successfully submitted
-	// error will be returned if maximum retries have been reached or the query to the consumer chain fails
+	// First iteration happens before the loop
 	for {
-		select {
-		case <-time.After(fp.cfg.SubmissionRetryInterval):
-			// error will be returned if max retries have been reached
-			var res *types.TxResponse
-			var err error
-			res, err = fp.SubmitBatchFinalitySignatures(targetBlocks)
-			if err != nil {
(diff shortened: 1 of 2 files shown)


finality-provider/service/fp_instance.go
```diff
@@ -403,58 +403,55 @@ func (fp *FinalityProviderInstance) retrySubmitSigsUntilFinalized(targetBlocks [
...
+			return nil, fmt.Errorf("failed to query block finalization at height %v: %w", targetHeight, err)
+		}
+		if finalized {
+			fp.logger.Debug(
+				"the block is already finalized, skip submission",
+				zap.String("pk", fp.GetBtcPkHex()),
+				zap.Uint64("target_height", targetHeight),
+			)
+			return nil, nil
+		}
```
lesterli, line 449: can we move these codes into the `case <-time.After(fp.cfg.SubmissionRetryInterval): `?

lesterli (approved):

finality-provider/service/fp_instance.go
bap2pecs: no, the idea is we want the logic to be executed in the **first** iteration

if we move it inside the case block, it will be executed after 300 seconds

bap2pecs:
will let @gitferry @SebastianElvis take a second look before I merge

gitferry (approved):
Nice find! Lgtm! Could you also back port it to main? Seems an important change

bap2pecs:
> Nice find! Lgtm! Could you also back port it to main? Seems an important change

https://github.com/babylonlabs-io/finality-provider/pull/227

Issue #194, opened 2024-12-05
Prune used merkle proof from db


gitferry:
Once a merkle proof is used after successful submission of finality vote, the merkle proof should be removed from db

PR #218, opened 2024-12-12
feat: prune used merkle proof

Lazar955:
[Closes](https://github.com/babylonlabs-io/finality-provider/issues/194)


Lazar955:
Hm, it's expected for double sign test to fail now, as the proof is deleted

gitferry (commented):
Keep thinking about this. I feel safer if we make it a CLI: `fpd unsafe-prune-merkle-proof [eots-pk] --chain-id --up-to-height` where `--chain-id` is required. If the user does not specify `--up-to-height`, the pruning is up to the fp's last voted height.

This would require db schema change as follows:
1. the current key is the randomness number itself, which does not support range query by height. Therefore we need to add height to the key
2. given that the fpd might store different fps with different chain id, we need also add chainid and fp public key in the key like we did for signing record https://github.com/babylonlabs-io/finality-provider/blob/27349c94ac8b5dc526813c71ad18bed67b8a95be/eotsmanager/store/storedsigningrecord.go#L22

wdyt?

bap2pecs:
> Keep thinking about this. I feel safer if we make it a CLI: `fpd unsafe-prune-merkle-proof [eots-pk] --chain-id --up-to-height` where `--chain-id` is required. If the user does not specify `--up-to-height`, the pruning is up to the fp's last voted height.
> 
> 
> 
> This would require db schema change as follows:
> 
> 1. the current key is the randomness number itself, which does not support range query by height. Therefore we need to add height to the key
> 
> 2. given that the fpd might store different fps with different chain id, we need also add chainid and fp public key in the key like we did for signing record https://github.com/babylonlabs-io/finality-provider/blob/27349c94ac8b5dc526813c71ad18bed67b8a95be/eotsmanager/store/storedsigningrecord.go#L22
> 
> 
> 
> wdyt?

i am not familiar with this db schema but the first point makes sense.

for 2), I don't know if it's necessary to change db because a FP can only register for one consumer chain. So we can simply query the FP's consumer chain id to see if it matches with what's passed in to the CLI

meta-thought: you guys might need to think about data migration for backward compatibility issue.

gitferry:
> for 2), I don't know if it's necessary to change db because a FP can only register for one consumer chain. So we can simply query the FP's consumer chain id to see if it matches with what's passed in to the CLI

We need to change db to add `height` as the key for each randomness along with the proof. However, a fpd may store randomness for different fps, so fp public key is also neede in the key. Probably we don't need chainid as the key for now as fp is unique across different chains but I think it's worth keeping it for potential range query for different chains

> meta-thought: you guys might need to think about data migration for backward compatibility issue.

my understanding for this is that after we launch phase-2, for any db schema change later on we need to consider backward compatibility. Do you think we need to do anything before the launch?

bap2pecs:
> my understanding for this is that after we launch phase-2, for any db schema change later on we need to consider backward compatibility. Do you think we need to do anything before the launch?

yeah i think it makes sense to only consider it after mainnet launch.

finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -488,6 +488,63 @@ func runCommandEditFinalityDescription(cmd *cobra.Command, args []string) error
 	return nil
 }
 
+// CommandUnsafeDeleteMerkleProof removes merkle proof
+func CommandUnsafeDeleteMerkleProof() *cobra.Command {
+	var cmd = &cobra.Command{
```
RafilxTenfen, line 493: should we have a command to delete this? kkk seems a bit scary

finality-provider/store/pub_rand.go
```diff
@@ -133,4 +171,37 @@ func (s *PubRandProofStore) GetPubRandProofList(pubRandList []*btcec.FieldVal) (
 	return proofBytesList, nil
 }
 
-// TODO: delete function?
+// RemovePubRandProofList removes all proofs up to the target height
+func (s *PubRandProofStore) RemovePubRandProofList(chainID []byte, pk []byte, targetHeight uint64) error {
```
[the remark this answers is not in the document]
Lazar955: thanks, added in [514b202](https://github.com/babylonlabs-io/finality-provider/pull/218/commits/514b202f139b7f21b7f5a7acf3e81b066b596f30)

finality-provider/cmd/fpd/daemon/daemon_commands.go
Lazar955: Yeah, maybe more docs about the consequences?

finality-provider/service/fp_instance.go
```diff
@@ -1020,3 +1026,8 @@ func (fp *FinalityProviderInstance) GetFinalityProviderSlashedOrJailedWithRetry(
 
 	return slashed, jailed, nil
 }
+
+// TestGetPubProof only used for tests to get access to the store
+func (fp *FinalityProviderInstance) TestGetPubProof(height uint64) ([]byte, error) {
```
gitferry, line 1031: This is probably not needed as we can access the store by `app.GetPubRandProofStore()`

Issue #196, opened 2024-12-05
Query consumer chain to check if the fp is already registered


gitferry:
If the fp is already registered, update local db to the remove finality provider info

PR #226, opened 2024-12-16
chore: query fp before register

Lazar955:
[Closes](https://github.com/babylonlabs-io/finality-provider/issues/196)

### DIFF
--- a/clientcontroller/interface.go
+++ b/clientcontroller/interface.go
@@ -51,6 +51,9 @@ type ClientController interface {
 		The following methods are queries to the consumer chain
 	*/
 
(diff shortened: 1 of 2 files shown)


finality-provider/service/app_test.go
gitferry, line 1: Can we have a test case where the fp is already registered but not in db, so calling `CreateFinalityProvider` will save the fp in db?

gitferry (approved):
Great work!

finality-provider/service/app_test.go
Lazar955: Done!

PR #227, opened 2024-12-16
backport: #217 - shouldn't wait SubmissionRetryInterval at the first iteration 

bap2pecs:
https://github.com/babylonlabs-io/finality-provider/pull/217/


Lazar955 (approved):

bap2pecs:
> Thanks! We probably need to do the same improve on the chain poller

good point. i just took a look and it seems there is the same problem. I won't have bandwidth to work on that, but feel free to create an issue to track.

```
func (cp *ChainPoller) pollChain() {
	defer cp.wg.Done()

	cp.waitForActivation()

	var failedCycles uint32

	for {
		select {
		case <-time.After(cp.cfg.PollInterval):
```

Issue #169, opened 2024-11-28
eotsd: Import key should save the keyname and eots pk mapping


gitferry:
This is needed to ensure that retrieving eots private key through pk could always work

PR #228, opened 2024-12-16
feat: add save key on mapping after import of keys

RafilxTenfen:
Closes: #169

### DIFF
--- a/eotsmanager/cmd/eotsd/daemon/keys.go
+++ b/eotsmanager/cmd/eotsd/daemon/keys.go
@@ -120,7 +120,7 @@ func saveKeyNameMapping(cmd *cobra.Command, keyName string) (*types.BIP340PubKey
 		return nil, fmt.Errorf("failed to save key name mapping: %w", err)
 	}
 
-	cmd.Printf("Sucessfully wrote key name %s to mapping", keyName)
+	cmd.Printf("Successfully wrote key name %s to mapping", keyName)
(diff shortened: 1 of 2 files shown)


eotsmanager/cmd/eotsd/daemon/keys.go
```diff
@@ -50,53 +50,84 @@ func NewKeysCmd() *cobra.Command {
...
+		eotsPk, err := saveKeyNameMapping(cmd, keyName)
+		if err != nil {
+			return err
+		}
+
+		return printFromKey(cmd, keyName, eotsPk)
 	}
 
+	saveKeyOnPostRun(keysCmd, "import")
+	saveKeyOnPostRun(keysCmd, "import-hex")
```
gitferry, line 63: let's also add `rename` here

gitferry (approved):
Nice work! Only comment is seems we miss `rename`

eotsmanager/cmd/eotsd/daemon/keys.go
RafilxTenfen: we currently have a check to avoid duplicated eots pk 
https://github.com/babylonlabs-io/finality-provider/blob/27349c94ac8b5dc526813c71ad18bed67b8a95be/eotsmanager/store/eotsstore.go#L65

I will merge this one as it is and we can think on how to solve this one

PR #231, opened 2024-12-17
fix: withdrawal cli for rewards fp

samricotta:
Currently there is an error when trying to set the withdrawal address this pr:

- fixes the error and allows to set an alternative withdrawal address
- fixes the formatting of the output when withdrawing

closes: https://github.com/babylonlabs-io/finality-provider/issues/353


gitferry (approved):
Nice work! Reminder is that please packport it to `release/v1.x`

gitferry (approved):
Please test it against testnet-5 before merge (might need ask devops to test it on an active fp). Also please ensure the operation doc is aligned

PR #233, opened 2024-12-18
chore: Refactor `CommitPubRand`

### LINKED ISSUE #225: Refactor CommitPubRand
CommitPubRand also includes saving all inclusion proofs of public randomness. This part should not be retried. We need to separate the function into:
1. determining the starting height to commit
2. generating pub rand and inclusion
3. committing public randomness.

gitferry:
Closes #225. The major goal of this pr is to refactor `CommitPubRand` but turned out included many small refactorings along the way (sorry for the mixup). Notable changes are:
- Split `retryCommitPubRandUntilBlockFinalized` into `ShouldCommitRandomness` and `CommitPubRand`. The former checks whether a new commit should be made based on the last committed height and the tip height. If `true`, it also returns the start height of the commit. The latter makes a new commit based on the start height returned by the former.
- Rename `config.MinRandHeightGap` to `config.TimestampingDelayBlocks` which is an estimated value of the latency between a commit is submitted and it is timestamped by the btc-timestamping protocol
- Remove status update loop. Instead, the status update is done when querying voting power during vote submission in `ProcessBlocksToVote`

Reviewers, please take a careful look at `fp_instance.go`, especially how `ShouldCommitRandomness` determines the start height with consideration of btc-timestamping delay


finality-provider/service/fp_instance.go
```diff
@@ -316,9 +373,17 @@ func (fp *FinalityProviderInstance) hasVotingPower(b *types.BlockInfo) (bool, er
 			zap.Uint64("block_height", b.Height),
 		)
 
+		if fp.GetStatus() == proto.FinalityProviderStatus_ACTIVE {
+			fp.MustSetStatus(proto.FinalityProviderStatus_INACTIVE)
```
Lazar955, line 377: Nit: not a huge fan of this, it's a bit hidden away that this function updates state. I suggest making it return the voting power or something like (bool, uint64, err), and then the function caller can decide what to do with it.

Alternatively, add comment saying that it also updates local status

finality-provider/service/fp_instance.go
```diff
@@ -173,41 +171,42 @@ func (fp *FinalityProviderInstance) finalitySigSubmissionLoop() {
 	defer fp.wg.Done()
 
 	for {
+		// start submission in the first iteration
+		pollerBlocks := fp.getAllBlocksFromChan()
```
Lazar955, line 175: nice, I bet we do this a lot throughout the codebases, that we wait before the first iteration

finality-provider/service/fp_instance.go
gitferry: Good point. Did some refactoring in [757156e](https://github.com/babylonlabs-io/finality-provider/pull/233/commits/757156e076588a39a50b42476ef87261648070d0)

finality-provider/service/fp_instance.go
gitferry: I just realized that we can't do it for every loop for cases where we have `continue` upon error. If we do, then we the loop will go forever. Need to figure out other ways for it

Lazar955 (approved):
Great work!

Issue #229, opened 2024-12-17
cli: Add `list` sub-command to `eotsd`


gitferry:
We need `list` sub-command in `eotsd` to show all the keyname <-> eots-pk mapping in the db. This helps operators to check if the key is successfully added/imported/recovered

Lazar955:
done in #234

PR #234, opened 2024-12-18
feat: list all keys in store

Lazar955:
Adds `eotsd ls` command which prints all keys from DB

[closes](https://github.com/babylonlabs-io/finality-provider/issues/229)

### DIFF
--- a/eotsmanager/cmd/eotsd/daemon/keys.go
+++ b/eotsmanager/cmd/eotsd/daemon/keys.go
@@ -128,10 +128,10 @@ func saveKeyNameMapping(cmd *cobra.Command, keyName string) (*types.BIP340PubKey
 // CommandPrintAllKeys prints all EOTS keys
 func CommandPrintAllKeys() *cobra.Command {
 	var cmd = &cobra.Command{
-		Use:     "ls",
+		Use:     "list",
 		Aliases: []string{"ls"},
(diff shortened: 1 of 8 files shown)


eotsmanager/cmd/eotsd/daemon/keys.go
```diff
@@ -124,6 +125,79 @@ func saveKeyNameMapping(cmd *cobra.Command, keyName string) (*types.BIP340PubKey
 	return eotsPk, nil
 }
 
+// CommandPrintAllKeys prints all EOTS keys
+func CommandPrintAllKeys() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:     "ls",
```
gitferry, line 131: this can be `list`?

eotsmanager/cmd/eotsd/daemon/keys.go
```diff
@@ -124,6 +125,79 @@ func saveKeyNameMapping(cmd *cobra.Command, keyName string) (*types.BIP340PubKey
...
+		Aliases: []string{"ls"},
+		Short:   "Print all EOTS keys from database",
```
gitferry, line 133:
```suggestion
		Short:   "Print all EOTS key names and public keys mapping from database",
```

gitferry (approved):
Great work!

Issue #243, opened 2024-12-22
EOTS key creation command timeouts


filippos47:
Version: `v0.14.1`

Key creation succeeds, but cmd fails with:

```shell
~ $ /bin/eotsd keys add finality-provider0 --keyring-backend=test --output=json
Error: failed to create db backend: timeout
```

The DB exists:

```shell
~/.eotsd $ tree
.
├── data
│   └── eots.db
├── eotsd.conf
├── keyring-test
│   ├── 24024d486963f86b60a52f37abbee00416eb1c04.address
│   └── finality-provider0.info
└── logs
    └── eotsd.log
```

gitferry:
I don't have problem doing it locally. @RafilxTenfen could you take a look as I will be off next Mon?

PR #244, opened 2024-12-22
fix: save key name mapping verifies if there is a client running

RafilxTenfen:
Closes: https://github.com/babylonlabs-io/finality-provider/issues/243


eotsmanager/cmd/eotsd/daemon/keys.go
```diff
@@ -92,6 +94,26 @@ func saveKeyNameMapping(cmd *cobra.Command, keyName string) (*types.BIP340PubKey
 		return nil, fmt.Errorf("failed to load config: %w", err)
 	}
 
+	client, err := eotsclient.NewEOTSManagerGRpcClient(cfg.RPCListener)
+	if err == nil {
```
KonradStaniec, line 98: I must say this seems like a lot guess work from the program it self. Isn't there a config option describing whether to used remote client or local db for key storage ?

RafilxTenfen: The thing here is if some other program has the db file open (`fpd start`) no other command can open the db to make modifications and it times out, so my idea here was check first if there is some client, updates by the client instead of trying to open the db file

KonradStaniec: what I am hinting at is that this behaviour of guessing who have the db open maybe a bit suprising and may have some unintended consequences. (like what if somebody have some invalid `cfg.RPCListener` in config and will now import key to some unexpected eots manager)

In my mind it should be explicit choice of the operator whether to import key locally through db directly or through running client and:
- in case of using import to db directly and db being open by other process, return error.
- in case of using running client, and client being not available, return error

RafilxTenfen: understood, what we could do is add an flag option to send to a running client

RafilxTenfen: updated

Lazar955 (approved):

KonradStaniec (approved):

PR #246, opened 2024-12-23
fix: start fp after register

Lazar955:
Starts the fp instance fp register


RafilxTenfen (approved):

RafilxTenfen:
integration broke :/

RafilxTenfen (approved):
LGTM :tada:

PR #255, opened 2024-12-27
Fix incorrect repository URL in finality provider operation guide

meetrick:
Corrected the repository URL
from https://github.com/babylonchain/finality-provider.git to https://github.com/babylonlabs-io/finality-provider.git

### DIFF
--- a/docs/finality-provider-operation.md
+++ b/docs/finality-provider-operation.md
@@ -74,8 +74,8 @@ Subsequently, clone the finality provider
 to the `bbn-test-5` tag. 
 <!-- TODO: change to a specific version -->
 
-```shell 
-git clone https://github.com/babylonchain/finality-provider.git
+```shell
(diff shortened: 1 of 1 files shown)


afa-farkhod:
it's also shown in the [main page](https://github.com/babylonchain)

image


```
BabylonChain Inc.
The Babylon project related repos are now mostly maintained under https://github.com/babylonlabs-io
```
so don't know whether it needs to be fixed through pr (generally no need to update) , that's obvious that you need to refer to another repo

meetrick:
> so don't know whether it needs to be fixed through pr (generally no need to update) , that's obvious that you need to refer to another repo

Your point is valid. However, if the guide is not appropriately updated, it could lead to confusion.

meetrick:
This issue is solved by https://github.com/babylonlabs-io/finality-provider/pull/256

PR #260, opened 2025-01-03
chore: Allow running of jailed fp

### LINKED ISSUE #259: FP should be able to start if jailed
We should allow `JAILED` fp to start in order to submit randomness so that randomness will be sufficient after unjailing

gitferry:
Closes #259. This allows jailed fps to still be able to send randomness and it could start voting immdiately after unjailing.


finality-provider/service/fp_instance.go
```diff
@@ -164,6 +165,15 @@ func (fp *FinalityProviderInstance) IsRunning() bool {
 }
 
 func (fp *FinalityProviderInstance) IsJailed() bool {
+	storedFp, err := fp.fpState.s.GetFinalityProvider(fp.GetBtcPk())
```
RafilxTenfen, line 168: it feels a bit weird to update the status here, but I see the reason

finality-provider/service/event_loops.go
```diff
@@ -189,6 +182,11 @@ func (app *FinalityProviderApp) unjailFpLoop() {
 				zap.String("txHash", res.TxHash),
 			)
 
+			// set the status to INACTIVE by default
```
RafilxTenfen, line 185: This is for the case where the fp was jailed, stoped the fpd, unjail in Babylon and start fpd again?

gitferry: I think the point of this pr is to allow the fp to be running continously running even though it got jailed and unjailed in the middle

finality-provider/service/fp_instance.go
gitferry: Yeah, I admit that this is a bit tricky. The status could be changed in the app level (the operator run `fpd unjail` while the fp is still running) if unjailing is processed so that the fp instance is not aware. Therefore, we need to check if the db has changed. Wondering if there's a better way to do it.

RafilxTenfen (approved):
LGTM :tada:

finality-provider/service/fp_instance.go
Lazar955: Let's add a godoc above the fcn

Lazar955 (approved):
lgtm

KonradStaniec (approved):

PR #262, opened 2025-01-05
feat: export pop command

RafilxTenfen:

PR #271, opened 2025-01-07
chore: backport #262 to v0.4.x and use old CLI

RafilxTenfen:


gitferry (commented):
Shouldn't go to `v0.14.x` and we release `v0.14.4`?

RafilxTenfen:
> Shouldn't go to `v0.14.x` and we release `v0.14.4`?

@vitsalis shared that we would like to be compatible with the finality providers that created in phase-1

gitferry (approved):

PR #277, opened 2025-01-10
feat: poll many blocks

### LINKED ISSUE #264: Improve poller to poll batch of blocks at once
Currently, the poller only polls a single block each time, which is troublesome during catch-up. It should be able to retrieve a batch during each poll

Lazar955:
closes #264

### DIFF
--- a/finality-provider/config/poller.go
+++ b/finality-provider/config/poller.go
@@ -1,6 +1,9 @@
 package config
 
-import "time"
+import (
+	"fmt"
+	"time"
+)
 
 var (
 	defaultBufferSize        = uint32(1000)
@@ -25,4 +28,16 @@ func DefaultChainPollerConfig() ChainPollerConfig {
 		StaticChainScanningStartHeight: defaultStaticStartHeight,
 		AutoChainScanningMode:          true,
 	}
+}
+
+func (c ChainPollerConfig) Validate() error {
+	if c.BufferSize == 0 {
+		return fmt.Errorf("invalid buffersize: %d", c.BufferSize)
+	}
+
+	if c.PollSize == 0 {
+		return fmt.Errorf("invalid pollinterval: %d", c.PollInterval)
+	}
+
+	return nil
 }
--- a/finality-provider/config/poller.go
+++ b/finality-provider/config/poller.go
@@ -6,6 +6,7 @@ var (
 	defaultBufferSize        = uint32(1000)
 	defaultPollingInterval   = 1 * time.Second
 	defaultStaticStartHeight = uint64(1)
+	defaultPollSize          = uint32(1000)
 )
 
 type ChainPollerConfig struct {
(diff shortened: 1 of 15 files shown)


finality-provider/service/chain_poller.go
```diff
@@ -261,9 +319,19 @@ func (cp *ChainPoller) SkipToHeight(height uint64) error {
...
 
 func (cp *ChainPoller) NextHeight() uint64 {
+	cp.mu.RLock()
+	defer cp.mu.RUnlock()
+
 	return cp.nextHeight
 }
 
+func (cp *ChainPoller) setNextHeight(height uint64) {
+	cp.mu.Lock()
```
Lazar955, line 329: if we run the unit tests with -race, we have a race condition, need a mutex for this

finality-provider/service/chain_poller.go
```diff
@@ -168,41 +202,65 @@ func (cp *ChainPoller) pollChain() {
...
-				zap.Uint64("height", block.Height))
-
-			// push the data to the channel
-			// Note: if the consumer is too slow -- the buffer is full
-			// the channel will block, and we will stop retrieving data from the node
-			cp.blockInfoChan <- block
+			// start polling in the first iteration
+			blockToRetrieve := cp.NextHeight()
+
+			blocks, err := cp.blocksWithRetry(blockToRetrieve, latestBlock.Height, latestBlock.Height)
```
gitferry, line 223: why the limit is set to `latestBlock.Height`? Maybe we need a config item for poller?

finality-provider/service/chain_poller.go
```diff
@@ -168,41 +202,65 @@ func (cp *ChainPoller) pollChain() {
...
-
-			cp.logger.Info("the poller retrieved the block from the consumer chain",
-				zap.Uint64("height", block.Height))
-
-			// push the data to the channel
-			// Note: if the consumer is too slow -- the buffer is full
-			// the channel will block, and we will stop retrieving data from the node
-			cp.blockInfoChan <- block
+			// start polling in the first iteration
+			blockToRetrieve := cp.NextHeight()
```
gitferry, line 221: I would add a small improvement here: if `blockToRetrieve == latestBlock.Height`, we don't need to call `cp.blocksWithRetry` but return []{latestBlock}. This will save a rpc call for the most common case

gitferry (commented):
Nice work! Two comments:

finality-provider/config/poller.go
```diff
@@ -6,19 +6,22 @@ var (
...
 	StaticChainScanningStartHeight uint64        `long:"staticchainscanningstartheight" description:"The static height from which we start polling the chain"`
 	AutoChainScanningMode          bool          `long:"autochainscanningmode" description:"Automatically discover the height from which to start polling the chain"`
+	PollSize                       uint32        `long:"pollsize" description:"The poll batch size when polling for blocks"`
 }
 
 func DefaultChainPollerConfig() ChainPollerConfig {
 	return ChainPollerConfig{
 		BufferSize:                     defaultBufferSize,
 		PollInterval:                   defaultPollingInterval,
+		PollSize:                       defaultPollSize,
```
gitferry, line 24: let's validate it in `cfg.Validate()`

gitferry (approved):
Nice work!

PR #283, opened 2025-01-14
docs: Update intro to include a quick start

samricotta:
A few users have been asking for a `quick start` or summary of what needs to happen. This includes a highlevel of what needs to be done so its the first thing people read

### DIFF
--- a/docs/finality-provider-operation.md
+++ b/docs/finality-provider-operation.md
@@ -1,7 +1,18 @@
 # Finality Provider Operation
 
-This document guides operators through the complete
-lifecycle of running a finality provider, including:
+## Quick Start
+
+Operating a finality provider on the Babylon network involves several key steps:
+
+1. Importing your EOTS keys from Phase-1 or creating new ones
+2. Starting the EOTS daemon
+3. Adding a finality provider key
+4. Configuring and starting the finality provider daemon
+5. Creating a finality provider
+
+## Overview
+
+This technical guide covers:
 
 * Installing and configuring the finality provider toolset
   (EOTS Manager and Finality Provider daemon)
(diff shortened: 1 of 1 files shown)


docs/finality-provider-operation.md
```diff
@@ -1,7 +1,18 @@
...
-lifecycle of running a finality provider, including:
+## Quick Start
+
+Operating a finality provider on the Babylon network involves several key steps:
+
+1. Importing your EOTS keys from Phase-1 or creating new ones
+2. Starting the EOTS daemon
+3. Adding a finality provider key
+4. Configuring and starting the finality provider daemon
+5. Creating a finality provider
```
vitsalis, line 11:
```suggestion
1. Operating an EOTS daemon program responsible for producing EOTS signatures
    by utilizing your EOTS key. If you participated in phase-1, you need to import your
    phase-1 EOTS key, otherwise you need to create one.
2. Setting up a finality provider daemon. The finality provider daemon will interact
    with the EOTS manager to create EOTS signatures and with the Babylon Chain
    to submit the signatures as transactions. The finality provider daemon requires
    a Babylon Chain account that is used to sign for the transactions.
3. Using the finality provider CLI to create your finality provider.
```

vitsalis (commented):
Wrote some improvement suggestions. However, I don't think we need a "quick start". I think in the "note for phase-1 finality providers" we can just outline that the core thing they need to do is import their keys and operate the software. We can add a detail about this in the overview as well.

docs/finality-provider-operation.md
vitsalis: Yep, makes more sense to just have the next session imo

PR #294, opened 2025-01-22
chore: Improve fpd start

### LINKED ISSUE #289: Improve `fpd start`
Current `fpd start` will not initiate the fp instance, which is confusing. To improve,
1. `fpd start` to start the fp if there's a registered fp record in db
2. ideally to have only one fp maintained in fpd, new registered fp can overwrite the old one

gitferry:
Closes #289. With this pr, the `fpd start` flow becomes the follows:

- If --eots-pk is specified, start the specified fp instance. Otherwise,
- if no fp records found from db, start the daemon and show logs asking for registration
- if there's one fp record found from db, start the daemon with the fp
- if there are multiple fp record found from db, fail the cmd execution asking for set of --eots-pk


finality-provider/cmd/fpd/daemon/daemon_commands.go
```diff
@@ -199,6 +199,8 @@ func runCommandCreateFP(ctx client.Context, cmd *cobra.Command, _ []string) erro
 
 	printRespJSON(res)
 
+	fmt.Println("You finality provider is successfully created. Please restart your fpd.")
```
Lazar955, line 202: let's use `cmd.Println`

Lazar955 (approved):
Nice work!

vitsalis:
Should we also update the docs to specify the rules for the `--eots-pk` flag?

Issue #282, opened 2025-01-14
bug: `commission-rate` required on `edit-finality-provider`


samricotta:
When running `edit-finality-provider` it requires you to enter the `commission-rate`, which isn't the desired functionality. If the `commission-rate` isn't present the following error will occur:

```
Whoops. There was an error while executing your fpd CLI 'failed to edit finality provider 6d3d0db52eaf19bf47e45ed367c75bcba440bf634c312c2e3223e47a4d81f16c err rpc error: code = Unknown desc = decimal string cannot be empty'%
```

We should follow Cosmos SDK’s [convention](https://github.com/cosmos/cosmos-sdk/blob/b0d20bce2b5ba335b7bfa702efba5e57b7a6a753/x/staking/keeper/msg_server.go#L160) which allows empty commission rate in the request.

PR #296, opened 2025-01-23
fix: edit finality provider `commission-rate`

samricotta:
closes: #282

Previously commission rate was required although it should be optional. This fix fetches the already set commission rate otherwise will submit the new one if it is specified by the operator.

### DIFF
--- a/finality-provider/service/client/rpcclient.go
+++ b/finality-provider/service/client/rpcclient.go
@@ -134,7 +134,6 @@ func (c *FinalityProviderServiceGRpcClient) QueryFinalityProviderInfo(ctx contex
 // EditFinalityProvider - edit the finality provider data.
 func (c *FinalityProviderServiceGRpcClient) EditFinalityProvider(
 	ctx context.Context, fpPk *bbntypes.BIP340PubKey, desc *proto.Description, rate string) error {
-
 	if rate == "" {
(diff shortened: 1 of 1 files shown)


finality-provider/service/client/rpcclient.go
```diff
@@ -134,10 +134,20 @@ func (c *FinalityProviderServiceGRpcClient) QueryFinalityProviderInfo(ctx contex
...
 	ctx context.Context, fpPk *bbntypes.BIP340PubKey, desc *proto.Description, rate string) error {
+	currentProvider, err := c.QueryFinalityProviderInfo(ctx, fpPk)
+	if err != nil {
+		return fmt.Errorf("failed to get current provider info: %w", err)
+	}
+
+	if rate == "" {
+		rate = currentProvider.FinalityProvider.Commission
+	}
+
```
Lazar955, line 145: rather, let's first check if rate == "" and do the  `c.QueryFinalityProviderInfo` inside the if statement

Lazar955 (approved):
Nice work!

gitferry (approved):
I'm OK with the current fix. But ideally we should fix it on Babylon side -- follow Cosmos SDK's convension where empty commission rate is accepted. However, I was worried that this would cause breaking change to state as this changes how some txs are handled. Maybe we can leave that to `v2`.

PR #297, opened 2025-01-23
feat: add new command to validate exported pop

RafilxTenfen:


Lazar955 (approved):

gitferry (commented):
Can we make it validate a JSON file? Better if we can make `pop export` to generate a JSON file

RafilxTenfen:
> Can we make it validate a JSON file? Better if we can make `pop export` to generate a JSON file

we could make it a file, but also with just a few things more in the shell it could direct the output to a file like `eotsd pop export > file_path` and on the validate `pop validate $(cat file_path)` 

Let me know if outputting to a file and reading from it is still needed, I will address in another PR

PR #301, opened 2025-01-24
chore: check if tx index enabled

Lazar955:


gitferry (commented):
I think this is specific to babylon so we should do the check when we initiate the babylon client, i.e., https://github.com/babylonlabs-io/finality-provider/blob/edc7f6845542611fb1ef6e1a1b2e9c49bcf3e0ef/clientcontroller/babylon.go#L46

KonradStaniec:
>I think this is specific to babylon so we should do the check when we initiate the babylon client, i.e.,

This is good point! 

Though one thing is, that functions starting with `New..` should be side effect free and querying external service is definitely counted as side effect. So if we want to that at client level we should introduce `func Start() error` which will perform this initialisation logic.

Lazar955:
> I think this is specific to babylon so we should do the check when we initiate the babylon client, i.e.,
> 
> https://github.com/babylonlabs-io/finality-provider/blob/edc7f6845542611fb1ef6e1a1b2e9c49bcf3e0ef/clientcontroller/babylon.go#L46

Make sense. Also agree with @KonradStaniec regarding `Start()`. Added in most recent commit

clientcontroller/interface.go
```diff
@@ -89,6 +91,8 @@ type ClientController interface {
 	// the value zero should be returned.
 	QueryFinalityActivationBlockHeight() (uint64, error)
 
+	NodeTxIndexEnabled() (bool, error)
```
gitferry, line 94: This should be removed

gitferry (approved):
Lgtm!

KonradStaniec (approved):

PR #327, opened 2025-02-09
fix: no add failed cycles count when chain poller no found new blocks

fyInALT:
When chain poller try to got new blocks, it will count a `failedCycles`, when it > `maxFailedCycles`, the finality-provider will exit by:

```go
		if failedCycles > maxFailedCycles {
			cp.logger.Fatal("the poller has reached the max failed cycles, exiting")
		}
```

BUT in chain poller, when the poller no got new blocks during `PollInterval`, it will also add `failedCycles`.

So when the chain no produce blocks in `PollInterval` * `maxFailedCycles`, the fp will exit.

For some chain, it may no produce blocks when no new tx, and in some times, the chain will pause by chain upgrade, so it will make the fp exit by the error.

So in this PR, we make `failedCycles ` no update when no new blocks.

### DIFF
--- a/finality-provider/service/chain_poller.go
+++ b/finality-provider/service/chain_poller.go
@@ -112,8 +112,14 @@ func (cp *ChainPoller) blocksWithRetry(start, end uint64, limit uint32) ([]*type
 			return err
 		}
 
+		// no need return error when just no found new blocks,
+		// the chain to poller may not produce new blocks.
 		if len(block) == 0 {
-			return fmt.Errorf("no blocks found for range %d-%d", start, end)
+			cp.logger.Warn(
+				"no blocks found for range",
+				zap.Uint64("start_height", start),
+				zap.Uint64("end_height", end),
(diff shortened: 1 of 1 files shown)


SebastianElvis:
Concept ACK.

If the poll request gives no new block and does not return an error, then this shall not be counted as a "failed" request. This is because in many situations there might be no new block upon a new poll request. For example,

- Poller runs too frequently
- The chain gets halted/apphash-ed
- The chain is under software upgrade
- Certain chains do not produce new block until there is new tx (in this context, Arbitrum Orbit chain)

Wdyt @KonradStaniec @gitferry @Lazar955 ?

SebastianElvis (approved):
Thanks for the contribution!

gitferry (commented):
Nice work. Very clean! Maybe the target could be `main` as `base` has been merged?

gitferry (approved):

PR #333, opened 2025-02-13
poller: skip if no new block is polled

SebastianElvis:
Part of https://github.com/babylonlabs-io/babylon-integration-deployment/pull/32

`latestBlockHeight` can be the current height if there is no new block during this polling iteration, but `blockToRetrieve` is set `cp.NextHeight()` , meaning that `blockToRetrieve` can be bigger than `latestBlockHeight` by `1`.

This PR adds a check on `blockToRetrieve > lastestBlockHeight`, and skips with some logging if this occurs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+### Improvements
+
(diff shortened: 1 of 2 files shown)


gitferry (approved):

gitferry:
Maybe a good chance to fix the [comment](https://github.com/babylonlabs-io/finality-provider/pull/327/files#r1953654557) here?

SebastianElvis:
> Maybe a good chance to fix the [comment](https://github.com/babylonlabs-io/finality-provider/pull/327/files#r1953654557) here?

Good idea fixed

PR #337, opened 2025-02-13
chore: Cleanup eots interfaces

gitferry:
Patially addressed https://github.com/babylonlabs-io/pm/issues/198 by removing removing `passphrase` from EOTS manager interface and some cleanups. Some notable changes:
- Removed `passphrase` parameter from every EOTS manager interface
- Removed `KeyRecord` interface as we don't rely on it to create EOTS keys anymore (eotsd keys add instead).


Lazar955 (approved):
Good job on the refactor

PR #351, opened 2025-02-18
feat: Add recover proof cmd

### LINKED ISSUE #293: Add randomness proof recover support
If fp db file is corrupted, we should have a way to recover the merkle proof. Currently, the only way user could do to recover merkle proof is to use `unsafe-commit-pub-rand` which is not easy to use and very unsafe in the sense that one needs to query the remote randomness commit history for a correct `start_height`. Otherwise, the the merkle proof is not compatible with the remote one.

We should add a helper cmd like `fpd recover-merkle-proof --start-height` to recover randomness proof in db for such cases. It should recover all the random proof from the `--start-height` to the last committed height on Babylon. If `--start-height` is not specified, it will recover proof for all the commit.

gitferry:
Closes #293. In particular, this PR:
- added a new cmd `recover-rand-proof --start-height` which checks the randomness commit on Babylon, re-generate randomness proof, and stores it in db.
- to check the randomness commit on babylon, a new query `QueryPublicRandCommitList` is added to babylon consumer client
- Note that this cmd currently only works for babylon consumer chain
- updated operational doc

### DIFF
--- a/clientcontroller/babylon/consumer.go
+++ b/clientcontroller/babylon/consumer.go
@@ -500,20 +500,16 @@ func (bc *BabylonConsumerController) QueryPublicRandCommitList(fpPk *btcec.Publi
 		if err := commit.Validate(); err != nil {
 			return nil, err
 		}
-		pagination.Key = res.Pagination.NextKey
 
-		commitStartHeight := commit.StartHeight
-		commitEndHeight := commit.EndHeight()
-
-		if startHeight > commitEndHeight {
-			continue
+		if startHeight <= commit.EndHeight() {
+			commitList = append(commitList, commit)
 		}
 
-		commitList = append(commitList, commit)
-
-		if startHeight >= commitStartHeight {
+		if res.Pagination == nil || res.Pagination.NextKey == nil {
 			break
 		}
+
+		pagination.Key = res.Pagination.NextKey
 	}
 
 	return commitList, nil
--- a/clientcontroller/babylon/consumer.go
+++ b/clientcontroller/babylon/consumer.go
@@ -463,3 +463,58 @@ func (bc *BabylonConsumerController) UnjailFinalityProvider(fpPk *btcec.PublicKe
 
 	return &types.TxResponse{TxHash: res.TxHash, Events: res.Events}, nil
 }
+
+// QueryPublicRandCommitList returns the public randomness commitments list from the startHeight to the last commit
+// the returned commits are ordered in the accenting order of the start height
+func (bc *BabylonConsumerController) QueryPublicRandCommitList(fpPk *btcec.PublicKey, startHeight uint64) ([]*types.PubRandCommit, error) {
+	fpBtcPk := bbntypes.NewBIP340PubKeyFromBTCPK(fpPk)
+
+	pagination := &sdkquery.PageRequest{
+		Limit:   1,
+		Reverse: false,
+	}
+
+	commitList := make([]*types.PubRandCommit, 0)
+
+	for {
+		res, err := bc.bbnClient.QueryClient.ListPubRandCommit(fpBtcPk.MarshalHex(), pagination)
+		if err != nil {
+			return nil, fmt.Errorf("failed to query committed public randomness: %w", err)
+		}
+
+		if len(res.PubRandCommitMap) == 0 {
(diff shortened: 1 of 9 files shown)


clientcontroller/babylon/consumer.go
```diff
@@ -463,3 +463,54 @@ func (bc *BabylonConsumerController) UnjailFinalityProvider(fpPk *btcec.PublicKe
...
 	return &types.TxResponse{TxHash: res.TxHash, Events: res.Events}, nil
 }
+
+// QueryPublicRandCommitList returns the public randomness commitments list from the startHeight to the last commit
+// the returned commits are ordered in the accenting order of the start height
+func (bc *BabylonConsumerController) QueryPublicRandCommitList(fpPk *btcec.PublicKey, startHeight uint64) ([]*types.PubRandCommit, error) {
+	fpBtcPk := bbntypes.NewBIP340PubKeyFromBTCPK(fpPk)
+
+	pagination := &sdkquery.PageRequest{
+		Limit:   1,
```
RafilxTenfen, line 473: why limit 1?

itest/babylon/babylon_e2e_test.go
```diff
@@ -1,5 +1,5 @@
-//go:build e2e_babylon
-// +build e2e_babylon
+//go:build e2e_op
+// +build e2e_op
```
RafilxTenfen, line 2: if we change the build tag here, it should modify in the Makefile as well

docs/finality-provider-operation.md
```diff
@@ -939,4 +943,43 @@ For Finality Provider:
...
+Babylon chain. Specifically, this can be achieved by repeating the
+[creation process](#61-create-finality-provider). The `create-finality-provider`
+cmd will download the info of the finality provider locally if it is already
+registered on Babylon.
+
+#### 7.3.2 Recover public randomness proof
+
+Every finality vote must contain the public randomness proof to prove that the
+randomness used in the signature is already committed on Babylon. Loss of
+public randomness proof leads to direct failure of the vote submission.
```
RafilxTenfen, line 978: We should add here at some point that the fpd must be stoped to run this command, otherwise it keeps the db locked

docs/finality-provider-operation.md
```diff
@@ -939,4 +943,43 @@ For Finality Provider:
...
 > 🔒 **Security Note**: While database files can be recreated, loss of private keys in the keyring directories is **irrecoverable** and will result in permanent loss of your finality provider position and accumulated rewards.
 
-> 💡 **Future Updates**: The ability to recreate the `fpd.db` from chain state will be offered in future releases, but keyring backup remains critical.
+### 7.3 Recover finality-provider.db
+
+The `finality-provider.db` file contains both the finality provider's running
+status and the public randomness merkle proof. Either information loss
+compromised will lead to service halt, but they are recoverable.
+
+#### 7.3.1 Recover local status of a finality provider
```
RafilxTenfen, line 952: If there is a loss in the db, the steps are

- Stop the fpd
- recover `fpd recover-rand-proof`
- Start fpd 
- Create finality provider `fpd create-finality-provider`
- Stop the fpd
- Start the fpd 

Is this expected?

RafilxTenfen (approved):
a few comments, but good job 
it worked like a charm

clientcontroller/babylon/consumer.go
gitferry: We can remove the limit but I thought with `1` it's cleaner as we don't need to iterator the commit map and sort them by the start height

itest/babylon/babylon_e2e_test.go
gitferry: Nice find! This is actually a mistake

docs/finality-provider-operation.md
gitferry: Good point!

docs/finality-provider-operation.md
gitferry: I think if the db file is lost, fpd cannot be started due to missing the fp object. So I think the first step should be start `fpd` and run `fpd create-finality-provider`. Then `stop fpd` and run `fpd recover-rand-proof`. Finaly, restart the `fpd`

Issue #352, opened 2025-02-19
Track finality provider rewards


gitferry:
The finality provider program should act as the dashboard for managing your finality provider. It should provide capabilities for querying the rewards and withdrawing them. Currently, the withdrawal cmd is there, lacking cmd for querying finality provider rewards. The query should point to [RewardGauges](https://github.com/babylonlabs-io/babylon/blob/5f8af8ced17d24f3f0c6172293cd37fb3d055807/client/query/incentive.go#L22) of the `incentive` module by the given BABY address

PR #356, opened 2025-02-25
chore(incentive): add reward gauges query cmd

GAtom22:
Closes https://github.com/babylonlabs-io/finality-provider/issues/352

### DIFF
--- a/docs/finality-provider-operation.md
+++ b/docs/finality-provider-operation.md
@@ -38,6 +38,7 @@ gain an overall understanding of the finality provider.
    4. [Slashing](#64-slashing)
    5. [Prometheus Metrics](#65-prometheus-metrics)
    6. [Withdrawing Rewards](#66-withdrawing-rewards)
+   7. [Query Rewards](#67-query-rewards)
 7. [Recover fpd.db](#7-recovery-and-backup)
    1. [Critical assets](#71-critical-assets)
    2. [Backup recommendations](#72-backup-recommendations)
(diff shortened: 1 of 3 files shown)


docs/finality-provider-operation.md
```diff
@@ -887,8 +888,20 @@ This will withdraw **ALL** accumulated rewards to the address you set in the
 `set-withdraw-addr` command if you set one. If no withdrawal address was set,
 the rewards will be withdrawn to your finality provider address.
 
+
```
gitferry, line 891: empty line

docs/finality-provider-operation.md
```diff
@@ -887,8 +888,20 @@ This will withdraw **ALL** accumulated rewards to the address you set in the
...
 Congratulations! You have successfully set up and operated a finality provider.
 
+### 6.7. Query Rewards
```
gitferry, line 894: I would put it as an optional step within 6.6 other than a new sub-section.

docs/finality-provider-operation.md
```diff
@@ -887,8 +888,20 @@ This will withdraw **ALL** accumulated rewards to the address you set in the
...
+
+You can query the reward gauges of a given stakeholder address.
+
+```shell
+fpd reward-gauges <address>
```
gitferry, line 899: Seems the `chain-id` flag is required? Better to add it here

gitferry (approved):
Lgtm! Minor comments on doc

docs/finality-provider-operation.md
GAtom22: I tested it without that flag and worked alright:

```
❯ fpd reward-gauges bbn128fsysa3c9wqlsc8l5y23jlsg8s708ph3pqahg --node https://rpc-dapp.devnet.babylonlabs.io/
reward_gauges:
  finality_provider:
    coins:
    - amount: "963863176"
      denom: ubbn
    withdrawn_coins: []
```
But can add it as optional anyways

GAtom22: Actually, when using the `chain-id` flag I get the error `unknown flag: --chain-id`

gitferry: Ah I see. Thanks for the confirmation! Maybe we should leave `--node` in the example given that it needs a connection to an RPC node

gitferry:
Also a kinder reminder to backport it to `release/v1.x` after merge as we do for the babylon repo

PR #367, opened 2025-03-06
Update finality-provider-operation.md

ghazni101:
added example commands for recovering finality-provider.db


gitferry:
Thanks! We have improved the doc in #391

PR #368, opened 2025-03-07
fix: fpd config validate

gitferry:
The issue only exists in `release/v1.x` branch


GAtom22 (approved):
LGTM! Thanks @gitferry!

PR #374, opened 2025-03-12
fix: check excluded methods with full names instead of prefixes and remove env variable requirement

Vvaradinov:
Check full proto names of the excluded from hmac methods instead of with suffix. Removes the requirement for ENV variable setting and only keeps config settings.


gitferry (approved):
Code wise lgtm. E2e tests failed though

Lazar955 (approved):

Issue #338, opened 2025-02-13
chore: EOTS key must enforce `--keyring-backend test`


samricotta:
When running the `EOTS` manager one needs to ensure they use `--keyring-backend test` for automatic signing of transactions. If this is not done there will be issues for the operators of FP's. This must be enforced in the backend.

PR #382, opened 2025-03-17
chore: enforcing `keyring-backend test` 

samricotta:
closes: https://github.com/babylonlabs-io/finality-provider/issues/338

Due to the need of continuous signing from the FP's we need to enforce `keyring-backend test` over file and os. This was seen as problematic for users during the testnet

### DIFF
--- a/eotsmanager/config/config.go
+++ b/eotsmanager/config/config.go
@@ -87,6 +87,10 @@ func (cfg *Config) Validate() error {
 		return fmt.Errorf("invalid RPC listener address %s, %w", cfg.RPCListener, err)
 	}
 
+	if cfg.KeyringBackend == "" {
+		return fmt.Errorf("the keyring backend should not be empty")
+	}
+
 	if cfg.Metrics == nil {
 		return fmt.Errorf("empty metrics config")
 	}
--- a/finality-provider/cmd/fpd/daemon/flags.go
+++ b/finality-provider/cmd/fpd/daemon/flags.go
@@ -11,7 +11,6 @@ const (
 	checkDoubleSignFlag  = "check-double-sign"
 	fromFile             = "from-file"
 	upToHeight           = "up-to-height"
-	keyringBackendFlag   = "keyring-backend"
 
 	// flags for description
 	monikerFlag         = "moniker"


eotsmanager/config/config.go
```diff
@@ -89,6 +89,8 @@ func (cfg *Config) Validate() error {
 
 	if cfg.KeyringBackend == "" {
 		return fmt.Errorf("the keyring backend should not be empty")
+	} else if cfg.KeyringBackend != "test" {
+		return fmt.Errorf("the keyring backend should be test")
```
Lazar955, line 93:
```suggestion
		return fmt.Errorf(`the keyring backend should be "test"`)
```

eotsmanager/config/config.go
```diff
@@ -89,6 +89,8 @@ func (cfg *Config) Validate() error {
 
 	if cfg.KeyringBackend == "" {
 		return fmt.Errorf("the keyring backend should not be empty")
+	} else if cfg.KeyringBackend != "test" {
```
gitferry, line 92: Seems we can just do `if cfg.KeyringBackend != "test"`?

gitferry (commented):
seems we also need to enforce it on `fpd`?

gitferry (commented):
Imo we should not enforce `test` in `keys add` cmd. Instead, we should do the check when the user starts the daemon (either fpd or eotsd). People may want to add a key with encryption but not used for operations

eotsmanager/config/config.go
```diff
@@ -87,8 +87,8 @@ func (cfg *Config) Validate() error {
 		return fmt.Errorf("invalid RPC listener address %s, %w", cfg.RPCListener, err)
 	}
 
-	if cfg.KeyringBackend == "" {
-		return fmt.Errorf("the keyring backend should not be empty")
+	if cfg.KeyringBackend != "test" {
+		return fmt.Errorf(`the keyring backend should be "test"`)
```
gitferry, line 91: let's do the `test` check in `eotsd start`

finality-provider/cmd/fpd/daemon/flags.go
```diff
@@ -11,13 +11,14 @@ const (
 	checkDoubleSignFlag  = "check-double-sign"
 	fromFile             = "from-file"
 	upToHeight           = "up-to-height"
+	keyringBackendFlag   = "keyring-backend"
```
gitferry, line 14: is this still needed?

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * [#331](https://github.com/babylonlabs-io/finality-provider/pull/331) Bump Cosmos integration dependencies
 * [#335](https://github.com/babylonlabs-io/finality-provider/pull/335) chore: fix CosmWasm controller
 * [#358](https://github.com/babylonlabs-io/finality-provider/pull/358) Bump Babylon dependency and add changes related to `x/btcstaking` module
+* [#382](https://github.com/babylonlabs-io/finality-provider/pull/382) chore: enforcing keyring-backend test
```
gitferry, line 49:
```suggestion
* [#382](https://github.com/babylonlabs-io/finality-provider/pull/382) chore: enforcing keyring-backend test in `fpd start` and `eotsd start`
```

finality-provider/cmd/fpd/daemon/start.go
```diff
@@ -59,6 +59,10 @@ func runStartCmd(ctx client.Context, cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("failed to load configuration: %w", err)
 	}
 
+	if cfg.BabylonConfig.KeyringBackend != "test" {
+		return fmt.Errorf("the keyring backend should be test")
```
gitferry, line 63:
```suggestion
		return fmt.Errorf("the keyring backend in config must be `test` for automatic signing, got %s", cfg.BabylonConfig.KeyringBackend)
```

gitferry (approved):
Thanks for the changes! Lgtm

PR #391, opened 2025-03-20
chore: Improve config and operation doc

### LINKED ISSUE #376: doc: Improve recover guideline
- --start-height not need to be specified. If set, should not be higher than the last submitted height
- Remove the FP file and restart the daemon
- Recreate the FP
- Stop the daemon and not start it again
- Run the recover command
- Start the daemon

### LINKED ISSUE #381: config: Should set higher default value for NumPubRand
Currently, 10k of NumPubRand only lasts for 1-day usage. Should set a higher value, e.g., 50k, to tolerate vigilante going down for 3-4 days. According to the bench result https://github.com/babylonlabs-io/finality-provider/pull/190, generating 50k rand take less than 1 min on M3 pro. The delay is OK compared with 10min BTC block time.

We should document the impact of `NumPubRand`, i.e., large value means better torlerace of down time of the fp and vigilante, but incurs higher cost both in gas fee (large size of proof) and storage, and let operators to carefully pick the value.

### LINKED ISSUE #385: config: set log level debug as default

gitferry:
Closes #385 
Closes #381 
Closes #376


finality-provider/proto/scripts/protocgen.sh
```diff
@@ -3,7 +3,6 @@
 set -eo pipefail
 
 cd proto
-buf dep update
```
gitferry, line 6: This change is due to 
```
❯ buf dep update
unknown command "dep" for "buf"
```
Any ideas?

finality-provider/proto/finality_providers.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.5
+// 	protoc-gen-go v1.31.0
```
gitferry, line 3: Not sure why this got downgraded

docs/finality-provider-operation.md
```diff
@@ -435,7 +443,7 @@ overwrite the config file they need to use `--force`.
 ├── config/
 │   └── fpd.conf       # Configuration file for the finality provider
 ├── data/
-│   └── fpd.db         # Database containing finality provider data
+│   └── finality-provider.db         # Database containing finality provider data
```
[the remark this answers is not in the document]
gitferry: The default db file is always `finality-provider.db` https://github.com/babylonlabs-io/finality-provider/blob/aae4976044c56acb64c98c3611605bb9b6b68d2c/finality-provider/config/db.go#L10. The doc was not consistent

RafilxTenfen (approved):
LGTM

PR #397, opened 2025-03-25
chore: overwrite randomness proof

### LINKED ISSUE #396: Should allow overwrite of randomness proof
Currently, the store will not overwrite randomness proof if the key already exists ([code ref](https://github.com/babylonlabs-io/finality-provider/blob/edc7f6845542611fb1ef6e1a1b2e9c49bcf3e0ef/finality-provider/store/pub_rand.go#L104)). This will cause issue if the proof is different than the previous one due to long down time (the commit range is different).

Lazar955:
closes #396


gitferry (approved):
Lgtm!

PR #401, opened 2025-03-25
chore: update deps

Lazar955:


KonradStaniec:
maybe you can also update Babylon node version to the latest snapshot ? https://github.com/babylonlabs-io/babylon/releases/tag/v1.99.0-snapshot.250319, this should fix a lot of security pings from dependabot

gitferry (approved):

PR #412, opened 2025-03-29
hotfix: Support two rand genrator for backward-compatibility

gitferry:
## Goal

For people who are running with old randomness generation scheme to commit with the new scheme so that they can switch to released version automatically after the old rand ran out.

## Approach

1. Fix will be implemented on the `hotfix` branch https://github.com/babylonlabs-io/finality-provider/tree/delete-sign-record-hotfix
2. Bump babylon dependency to rc.8
3. Implement 2 versions of rand generating scheme in the codebase
4. Always use the new scheme to generate new commit, and store the proof
5. Add a new api `SignEOTSBoth` to generate 2 signatures along with randomness, one using the old scheme, the other one using the new scheme
6. The fpd will pick one pair to submit sigs according to proof verification
7. For anti-slashing, no longer save the signature. If the same app hash is requested, generate the sig. Otherwise, raise an alarm

## TODOs
- Support two schemes when recover randomness


eotsmanager/randgenerator/randgenerator.go
```diff
@@ -13,7 +13,41 @@ import (
...
+
+	return &privRand.Key, &j.X
+}
+
+// GenerateRandomnessUnsafe generates a random scalar with the given key and src
+// the result is deterministic with each given input
+// NOTE: keeping it is for backward compatibility
+//
+//	this will be deprecated soon
+func GenerateRandomnessUnsafe(key []byte, chainID []byte, height uint64) (*eots.PrivateRand, *eots.PublicRand) {
```
SebastianElvis, line 49: How about naming this `GenerateRandomnessLegacy` rather than unsafe? as

- There is only a small bias in the legacy rand generator, and we don't have an exploit for it. So it's a bit too harsh to call it "unsafe"
- It might be a bit confusing with "unsafe" commands that potentially lead to slashing. Using this rand generator won't lead to slashing

finality-provider/service/fp_instance.go
```diff
@@ -717,6 +718,126 @@ func (fp *FinalityProviderInstance) SubmitBatchFinalitySignatures(blocks []*type
...
+	// Try unsafe version first
+	res, err := fp.trySignatureVersion(block, sigRecordUnsafe.PubRand, proof, sigRecordUnsafe.Sig, "unsafe")
+	if err == nil {
+		return res, nil
+	}
+
+	res, err = fp.trySignatureVersion(block, sigRecord.PubRand, proof, sigRecord.Sig, "safe")
+	if err != nil {
+		return nil, fmt.Errorf("both signature versions failed: %w", err)
+	}
```
SebastianElvis, line 785: Is it possible to implement something as follows:

- query the pub rand at that height
- get the sign record with the same pub rand
- only submit the finality sig in that sign record

This way we avoid the retry completely

eotsmanager/localmanager.go
```diff
@@ -247,6 +247,73 @@ func (lm *LocalEOTSManager) SignEOTS(eotsPk []byte, chainID []byte, msg []byte,
...
+		Sig:     sigUnsafe,
+		PubRand: pubRandUnsafe,
+	}
+
+	// save the sign record as the placeholder
+	// note that only `msg` matters
+	s := sig.Bytes()
+	if err := lm.es.SaveSignRecord(height, chainID, msg, eotsPk, s[:]); err != nil {
+		return nil, nil, fmt.Errorf("failed to save signing record: %w", err)
+	}
```
SebastianElvis, line 312: Do we still need to have `s[:]` in the sign record KV store?

SebastianElvis (commented):
Some quick comments, overall great work :+1:

eotsmanager/localmanager.go
gitferry: only `msg` matters, `s` is just for placeholder

eotsmanager/randgenerator/randgenerator.go
gitferry: Make sense!

eotsmanager/localmanager.go
SebastianElvis: then shall we refactor `SaveSignRecord` to remove the parameter `s[:]`? could be done in future PRs though

finality-provider/service/fp_instance.go
gitferry: Good point. I just realized that we can use the proof to verify which rand to use without any query

eotsmanager/localmanager.go
gitferry: Not sure we should do it in `main` branch as one point of anti-slashing is never regenerate sig for the same height. Think we can revisit this refactor in `main`

Lazar955 (commented):
solid work!

KonradStaniec (approved):

eotsmanager/client/rpcclient.go
```diff
@@ -118,6 +119,43 @@ func (c *EOTSManagerGRpcClient) SignEOTS(uid, chaiID, msg []byte, height uint64)
...
 }
 
+func (c *EOTSManagerGRpcClient) SignEOTSBoth(uid, chaiID, msg []byte, height uint64) (*types.EOTSRecord, *types.EOTSRecord, error) {
+	req := &proto.SignEOTSBothRequest{
+		Uid:     uid,
+		ChainId: chaiID,
+		Msg:     msg,
+		Height:  height,
+	}
+	res, err := c.client.SignEOTSBoth(context.Background(), req)
```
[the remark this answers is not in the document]
gitferry: Good point! Seems a general improvement. Could you create an issue about it?

PR #417, opened 2025-03-31
chore: Modify default config and add validation

gitferry:
The point of this PR is to ensure that users cannot insert config values that will cause the fp to crash or jailed. Notable changes are:
- Remove config items `NumPubRandMax`, `TimestampingDelayBlocks`, `RandomnessCommitInterval` but hardcode them as they are critical ones that should not be changeable. `TimestampingDelayBlocks` is calculated based on `w=300` and babylon block production time of `10s`
- Add validations to config items. Note that `MinPubRand` to ensure that a commit can pass verification on Babylon side.

@babylonlabs-io/integration-dev Not sure this will have impact on the integration side. Please review


gitferry:
Just realized that the config change may not make sense to integration. So closing this and raised another PR targeting `release/v1.x` #418

PR #418, opened 2025-03-31
chore: Modify default config and add validation

gitferry:
The point of this PR is to ensure that users cannot insert config values that will cause the fp to crash or jailed. Notable changes are:

- Remove config items NumPubRandMax, TimestampingDelayBlocks, RandomnessCommitInterval but hardcode them as they are critical ones that should not be changeable. TimestampingDelayBlocks is calculated based on w=300 and babylon block production time of 10s
- Add validations to config items. Note that MinPubRand to ensure that a commit can pass verification on Babylon side.

### DIFF
--- a/finality-provider/config/config.go
+++ b/finality-provider/config/config.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"fmt"
-	"net"
 	"path/filepath"
 	"strconv"
 	"time"
@@ -17,6 +16,7 @@ import (
 	"github.com/babylonlabs-io/finality-provider/util"
 )
 
+// Constants for config default values
 const (
 	defaultChainType                   = "babylon"
 	defaultLogLevel                    = zapcore.DebugLevel
@@ -26,15 +26,21 @@ const (
 	DefaultRPCPort                     = 12581
 	defaultConfigFileName              = "fpd.conf"
 	defaultNumPubRand                  = 50000 // support running of roughly 5 days with block production time as 10s
-	defaultNumPubRandMax               = 500000
-	defaultTimestampingDelayBlocks     = 6000 // 100 BTC blocks * 600s / 10s
-	defaultBatchSubmissionSize         = 1000
-	defaultRandomInterval              = 30 * time.Second
+	defaultBatchSubmissionSize         = 100
 	defaultSubmitRetryInterval         = 1 * time.Second
 	defaultSignatureSubmissionInterval = 1 * time.Second
 	defaultMaxSubmissionRetries        = 20
 	defaultBitcoinNetwork              = "signet"
 	defaultDataDirname                 = "data"
+)
+
+// Constants for system parameters validation limits
+const (
+	MaxBatchSize            = 100              // Maximum allowed batch submission size
+	MaxPubRand              = 500000           // Maximum allowed public randomness number
+	MinPubRand              = 8192             // Minimum allowed public randomness number
+	TimestampingDelayBlocks = 18000            // 300 BTC blocks * 600s / 10s where 300 BTC blocks is the system parameter of BTC block time required by the btc timestamping protocol
+	RandCommitInterval      = 30 * time.Second // Interval between check of randomness commit
 )
 
 var (
@@ -55,15 +61,15 @@ type Config struct {
 	// ChainType and ChainID (if any) of the chain config identify a consumer chain
 	ChainType                   string        `long:"chaintype" description:"the type of the consumer chain" choice:"babylon"`
 	NumPubRand                  uint32        `long:"numPubRand" description:"The number of Schnorr public randomness for each commitment"`
-	NumPubRandMax               uint32        `long:"numpubrandmax" description:"The upper bound of the number of Schnorr public randomness for each commitment"`
-	TimestampingDelayBlocks     uint32        `long:"timestampingdelayblocks" description:"The delay, measured in blocks, between a randomness commit submission and the randomness is BTC-timestamped"`
 	MaxSubmissionRetries        uint32        `long:"maxsubmissionretries" description:"The maximum number of retries to submit finality signature or public randomness"`
 	EOTSManagerAddress          string        `long:"eotsmanageraddress" description:"The address of the remote EOTS manager; Empty if the EOTS manager is running locally"`
 	HMACKey                     string        `long:"hmackey" description:"The HMAC key for authentication with EOTSD. If not provided, will use HMAC_KEY environment variable."`
 	BatchSubmissionSize         uint32        `long:"batchsubmissionsize" description:"The size of a batch in one submission"`
-	RandomnessCommitInterval    time.Duration `long:"randomnesscommitinterval" description:"The interval between each attempt to commit public randomness"`
 	SubmissionRetryInterval     time.Duration `long:"submissionretryinterval" description:"The interval between each attempt to submit finality signature or public randomness after a failure"`
 	SignatureSubmissionInterval time.Duration `long:"signaturesubmissioninterval" description:"The interval between each finality signature(s) submission"`
+
+	// not configurable in config file
+	TimestampingDelayBlocks uint32
 
 	BitcoinNetwork string `long:"bitcoinnetwork" description:"Bitcoin network to run on" choice:"mainnet" choice:"regtest" choice:"testnet" choice:"simnet" choice:"signet"`
 
@@ -91,11 +97,9 @@ func DefaultConfigWithHome(homePath string) Config {
 		DatabaseConfig:              DefaultDBConfigWithHomePath(homePath),
 		BabylonConfig:               &bbnCfg,
 		PollerConfig:                &pollerCfg,
+		TimestampingDelayBlocks:     TimestampingDelayBlocks,
 		NumPubRand:                  defaultNumPubRand,
-		NumPubRandMax:               defaultNumPubRandMax,
-		TimestampingDelayBlocks:     defaultTimestampingDelayBlocks,
 		BatchSubmissionSize:         defaultBatchSubmissionSize,
-		RandomnessCommitInterval:    defaultRandomInterval,
 		SubmissionRetryInterval:     defaultSubmitRetryInterval,
 		SignatureSubmissionInterval: defaultSignatureSubmissionInterval,
 		MaxSubmissionRetries:        defaultMaxSubmissionRetries,
@@ -158,6 +162,8 @@ func LoadConfig(homePath string) (*Config, error) {
 		return nil, err
 	}
 
+	cfg.TimestampingDelayBlocks = TimestampingDelayBlocks
+
 	// Make sure everything we just loaded makes sense.
 	if err := cfg.Validate(); err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


finality-provider/config/config_test.go
```diff
@@ -0,0 +1,143 @@
...
+				MaxSubmissionRetries:        50,
+				NumPubRand:                  config.MinPubRand,
+				PollerConfig:                nil,
+			},
+			wantErr: "poller config cannot be empty",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
Lazar955, line 120:
```suggestion
		
```

Lazar955 (approved):
Overall looks good. Make sense to reduce config options especially the critical ones that could break fp

vitsalis:
Some questions:
* Should we add notes to the changelog or the release notes about the reduction in configuration parameters?
* It might be nice for a future addition to automatically calculate `TimestampingDelayBlocks` instead of hardcoding it. For example, in the testnet, we have `w=100`, so this config value will be wrong.

finality-provider/config/config.go
```diff
@@ -26,17 +26,23 @@ const (
...
 	defaultBitcoinNetwork              = "signet"
 	defaultDataDirname                 = "data"
 )
 
+// Constants for system parameters validation limits
+const (
+	MaxBatchSize            = 100              // Maximum allowed batch submission size
+	MaxPubRand              = 500000           // Maximum allowed public randomness number
+	MinPubRand              = 8192             // Minimum allowed public randomness number
+	TimestampingDelayBlocks = 18000            // 300 BTC blocks * 600s / 10s where 300 BTC blocks is the system parameter of BTC block time required by the btc timestamping protocol
```
vitsalis, line 42: What's the reasoning behind the `600s / 10s` part?

gitferry:
> Should we add notes to the changelog or the release notes about the reduction in configuration parameters?

Good idea. Will do in the next release

> It might be nice for a future addition to automatically calculate TimestampingDelayBlocks instead of hardcoding it. For example, in the testnet, we have w=100, so this config value will be wrong.

TimestampingDelayBlocks is calculated based on `w` from babylon and block production time. The former can actually be loaded into memory upon initiation, while the latter can remain a config value. Wdty?

finality-provider/config/config.go
gitferry: 10s is the babylon block production time. TimestampingDelayBlocks basically converts `w` into babylon blocks

KonradStaniec, line 42: hmm this value is incorrect for testnet ? does it mean new realease with this hardcoded won't be compatible with testnet ? or this value is compatible with both mainnet and testnet, and on testnet it will lead to committing randomness with more advance?

vitsalis:
Do we also need to update `docs/commit-pub-rand.md`?

finality-provider/config/config.go
vitsalis: How about we increase this to account for some buffer? I wouldn't even mind doubling it.

finality-provider/config/config.go
```diff
@@ -26,17 +26,23 @@ const (
...
 	defaultMaxSubmissionRetries        = 20
 	defaultBitcoinNetwork              = "signet"
 	defaultDataDirname                 = "data"
 )
 
+// Constants for system parameters validation limits
+const (
+	MaxBatchSize            = 100              // Maximum allowed batch submission size
+	MaxPubRand              = 500000           // Maximum allowed public randomness number
+	MinPubRand              = 8192             // Minimum allowed public randomness number
```
vitsalis, line 41: Let's add comments explaining the reasoning behind the selection of the above? I'm not sure I understand the selections.

finality-provider/config/config.go
```diff
@@ -26,17 +26,23 @@ const (
...
+	TimestampingDelayBlocks = 18000            // 300 BTC blocks * 600s / 10s where 300 BTC blocks is the system parameter of BTC block time required by the btc timestamping protocol
+	RandCommitInterval      = 30 * time.Second // Interval between check of randomness commit
```
vitsalis, line 43: Why do we choose this value?

finality-provider/config/config.go
```diff
@@ -55,16 +61,16 @@ type Config struct {
...
 	EOTSManagerAddress          string        `long:"eotsmanageraddress" description:"The address of the remote EOTS manager; Empty if the EOTS manager is running locally"`
 	HMACKey                     string        `long:"hmackey" description:"The HMAC key for authentication with EOTSD. If not provided, will use HMAC_KEY environment variable."`
 	BatchSubmissionSize         uint32        `long:"batchsubmissionsize" description:"The size of a batch in one submission"`
-	RandomnessCommitInterval    time.Duration `long:"randomnesscommitinterval" description:"The interval between each attempt to commit public randomness"`
 	SubmissionRetryInterval     time.Duration `long:"submissionretryinterval" description:"The interval between each attempt to submit finality signature or public randomness after a failure"`
 	SignatureSubmissionInterval time.Duration `long:"signaturesubmissioninterval" description:"The interval between each finality signature(s) submission"`
 
+	// not configurable in config file
+	TimestampingDelayBlocks uint32
```
vitsalis, line 72: Why add it here instead of loading the default from the package directly as you do for `RandCommitInterval, `MinNumPubRand` and so on?

finality-provider/config/config.go
```diff
@@ -158,6 +162,8 @@ func LoadConfig(homePath string) (*Config, error) {
 		return nil, err
 	}
 
+	cfg.TimestampingDelayBlocks = TimestampingDelayBlocks
```
vitsalis, line 165: Continuation on the above on whether this is needed.

finality-provider/config/config.go
gitferry: Sorry, the commens are not clear. This is the poll interval between each check of whether a new commit needs to be made. Technically, it depends on the value of `NumPubRand` and could be a large value but think we can hard-code a small value for safe

finality-provider/config/config.go
gitferry: This is because we need to set a small value for e2e test to pass quickly

gitferry:
> Also, should the base of this branch be main and then be backported to the release?

I was worried that some values do not make sense to phase-3. For example, `TimestampingDelayBlocks` is calculated based on babylon production time. For OP-Stack, the time could be shorter. We probably need a bigger refactoring for `main`

finality-provider/config/config.go
gitferry: Increasing it can reduce the number of wasted randomness due to timestamping but also delay the time of an fp becoming active. Given that the fp will have 2 available commits in the air, I think it's fine. Maybe can increase it a little bit like `20000`?

gitferry: I think it just makes the delay a bit longer (a new fp needs more time to become active) but still works for both testnet and mainnet. We can further modify it to query `w` from Babylon during initiation

finality-provider/config/config.go
KonradStaniec: hmm given that this poses some problems for:
- e2e tests
- switching between testnet / mainnet
- potential problems for phase-3 (which we went around by making pr to `release/v1.x`)

Maybe lets leave it as config value with default set to mainnet value ? and in future pr's introduce validation based on `W` from babylon node or even set it based on `W` from Babylon node

gitferry: No problem with me. I can add the value for testnet in the commends. Do we have a consensus on the default value for both mainnet and test https://github.com/babylonlabs-io/finality-provider/pull/418#discussion_r2020852928?

finality-provider/config/config.go
```diff
@@ -26,15 +26,36 @@ const (
...
+	defaultBatchSubmissionSize         = 100
 	defaultSubmitRetryInterval         = 1 * time.Second
 	defaultSignatureSubmissionInterval = 1 * time.Second
 	defaultMaxSubmissionRetries        = 20
 	defaultBitcoinNetwork              = "signet"
 	defaultDataDirname                 = "data"
+	// TimestampingDelayBlocks is the delay for a commit to become available due to BTC timestamping protocol
+	// it is calculated by converting BTC block time (parameter w=300) to Babylon Genesis block time
+	// 300 BTC blocks * 600s / 10s where 300 BTC blocks is the system parameter of BTC block time required
+	// by the btc timestamping protocol. Adding 12,000 as a buffer in case 300 BTC blocks come with more than
```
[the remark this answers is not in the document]
gitferry: Ops. Seems buffer is not needed after offline discussion. Removed it in the comments

KonradStaniec (approved):

finality-provider/config/config.go
```diff
@@ -26,15 +26,35 @@ const (
...
 	defaultSubmitRetryInterval         = 1 * time.Second
 	defaultSignatureSubmissionInterval = 1 * time.Second
 	defaultMaxSubmissionRetries        = 20
 	defaultBitcoinNetwork              = "signet"
 	defaultDataDirname                 = "data"
+	// TimestampingDelayBlocks is the delay for a commit to become available due to BTC timestamping protocol
+	// it is calculated by converting BTC block time (parameter w=300) to Babylon Genesis block time
+	// 300 BTC blocks * 600s / 10s where 300 BTC blocks is the system parameter of BTC block time required
+	// by the btc timestamping protocol.
+	// for testnet with w=100, the recommended default value is 6000
```
vitsalis, line 39:
```suggestion
	// TimestampingDelayBlocks is an estimation on the delay
	// for a commit to become available due to BTC timestamping protocol,
	// counted as the number of Babylon Genesis' blocks.
	// It is calculated by converting the time it takes for the public randomness
	// to be BTC timestamped counted as Bitcoin blocks to Babylon Genesis blocks.
	// As this is an estimation, we make the following assumptions:
	//    - Babylon Genesis' block time: 10s
	//    - Bitcoin's block time: 10m => 60 Babylon Genesis blocks per Bitcoin block.
	//    - Babylon Genesis' finalization delta: 300 blocks
	// The above leads to a recommended default value of 300 * 60 = 18000 Babylon blocks.
	// You should set this parameter depending on the network you connect to.
```

finality-provider/config/config.go
```diff
@@ -26,15 +26,35 @@ const (
...
+	defaultTimestampingDelayBlocks = 18000
+)
+
+// Constants for system parameters validation limits
+const (
+	// MaxBatchSize is the maximum allowed batch submission size,
+	// this limits BatchSubmissionSize as larger values could cause error due to insufficient fees
```
vitsalis, line 46:
```suggestion
	// MaxBatchSize is the maximum allowed batch submission size.
	// This constant BatchSubmissionSize as larger values could cause an error due to insufficient fees.
```

finality-provider/config/config.go
```diff
@@ -26,15 +26,35 @@ const (
...
+	MaxBatchSize = 100
+	// MaxPubRand is the maximum allowed number of public randomness in each commit,
+	// this limits NumPubRand as larger values could take more than 10min to generate/store
+	// rand proofs
```
vitsalis, line 50:
```suggestion
	// MaxPubRand is the maximum allowed number of public randomness in each public randomness commit.
	// This constant limits NumPubRand as larger values could take more than 10min to generate/store
	// the randomness proofs.
```

finality-provider/config/config.go
```diff
@@ -26,15 +26,35 @@ const (
...
+	MaxPubRand = 500000 //
+	// MinPubRand is the minimum allowed number of public randomness in each commit
+	// a commit with less than this value will be rejected by Babylon Genesis
```
vitsalis, line 53:
```suggestion
Suggested change
	// MinPubRand is the minimum allowed number of public randomness in each public randomness commit.
	// A commit containing a smaller amount of randomness might be rejected by Babylon Genesis.
```

vitsalis: > A commit containing a smaller amount of randomness might be rejected by Babylon Genesis.

Although this depends on the network

finality-provider/config/config.go
```diff
@@ -26,15 +26,35 @@ const (
...
+	MinPubRand = 8192
```
vitsalis, line 54: I think we should increase this. `8192` lasts only for ~22 hours, which would be below the required threshold for mainnet. We could set it to something like 18k or 16384

finality-provider/config/config.go
```diff
@@ -26,15 +26,35 @@ const (
...
+	// RandCommitInterval is the interval between each check of whether a new commit needs to be made,
+	// technically this could be a large value depending on NumPubRand, but hardcode a small value for safety
+	RandCommitInterval = 30 * time.Second // Interval between check of randomness commit
```
vitsalis, line 57:
```suggestion
	RandCommitInterval = 30 * time.Second
```

vitsalis (approved):

finality-provider/config/config.go
KonradStaniec: lets set it to 18k  👍

KonradStaniec: after second thought, maybe `MinPubRand` should be somehow based on `TimestampingDelayBlocks` 🤔 It seems the larger `TimestampingDelayBlocks` the larger `MinPubRand` should be.

gitferry: I set it to 18k and merging this PR.

> after second thought, maybe MinPubRand should be somehow based on TimestampingDelayBlocks 🤔 It seems the larger TimestampingDelayBlocks the larger MinPubRand should be.

I don't see a clear relation like this. I think as long as `NumPubRand` is large enough to cover the inaccuracy (btc being slow and randomness might run out when the commit is timestamped) of btc time estimation is OK.

PR #424, opened 2025-04-02
chore: Bump babylon v1.0.0

gitferry:
To release rc6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+## v1.0.0-rc.6
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+## v1.0.0-rc.6
```
RafilxTenfen, line 40: should we keep rc here? or release v1 as well :eyes:

RafilxTenfen (approved):

CHANGELOG.md
KonradStaniec: Ideally the switch between last rc and final release does not have any changes to signal that rc becomes release. So I would release rc.6 and then release v1 afterwards.

PR #427, opened 2025-04-02
fix: handle uninitialized commissionInfo when editing fp

Lazar955:
closes #425

PR #430, opened 2025-04-04
fix: handle uninitialized commissionInfo when editing fp (#427) [backport]

Lazar955:


Lazar955:
@gitferry, @maiquanghiep tested on testnet, fix works

gitferry (approved):

PR #450, opened 2025-04-11
chore: metrics for votes

Lazar955:
Introduces a new metric `fp_voted_height` so we can query historically voted heights.
func `RecordFpVoteTime` wasn't getting called, and it's important for `fp_seconds_since_last_vote`


gitferry (approved):
Nice work!

PR #454, opened 2025-04-18
Update finality provider docs

gitferry:
This PR improves the finality provider doc in general. It covers the following changes:
- Rename Babylon, Babylon network, Babylon blockchain to Babylon Genesis
- Improve `Rewards and Refunding`
  - Remove Refunding part
  - Add `6.3. Set Up Operation Key` to avoid confusion
  - Other small edits

### DIFF
--- a/docs/finality-provider-operation.md
+++ b/docs/finality-provider-operation.md
@@ -1,17 +1,16 @@
 # Finality Provider Operation
 
-This document guides operators through the complete
+This is an operational guide intended for technical finality provider administrators.
+This guide covers the complete
 lifecycle of running a finality provider, including:
 
 * Installing and configuring the finality provider toolset
-  (EOTS Manager and Finality Provider daemon)
-* Managing keys (EOTS key for signatures and Babylon key for rewards)
-* Registering your finality provider on the Babylon network
-* Operating and maintaining your finality provider
-* Collecting rewards
+  (EOTS Manager and Finality Provider daemon).
+* Managing keys (EOTS key for EOTS signatures and Babylon key for rewards).
+* Registering finality providers on Babylon Genesis.
+* Operating finality provider.
+* Claiming rewards.
 
-This is an operational guide intended for technical finality provider administrators.
-For conceptual understanding, see our [Technical Documentation](./fp-core.md).
 Please review the [high-level explainer](../README.md) before proceeding to
 gain an overall understanding of the finality provider.
 
@@ -49,24 +48,24 @@ gain an overall understanding of the finality provider.
 
 ## 1. A note about Phase-1 Finality Providers
 
-Thank you for participating in the first phase of the Babylon launch. This guide
-provides instructions for setting up the full finality provider toolset required
-for your participation in the second phase of the Babylon launch.
+This node is for you if you have participated in Phase-1 of Babylon
+Genesis to help you transition to Phase-2.
 
-Finality providers that received delegations on the first phase of the launch
-are required to transition their finality providers to the second phase
+Finality providers that received delegations in Phase-1
+are required to transition their finality providers to Phase-2
 using the same EOTS key that they used and registered with during Phase-1.
 The usage of a different key corresponds to setting up an entirely
 different finality provider which will not inherit the Phase-1 delegations.
-Not transitioning your Phase-1 finality provider prevents your Phase-1 delegations
-from transitioning to the second phase.
+If you don't transition your Phase-1 finality provider to Phase-2, all the
+received Phase-1 delegations cannot transition to Phase-2.
 
-If you already have set up a key during Phase-1, please proceed to
-[Adding Keys](#42-add-an-eots-key) to import your Phase-1 key.
+If you already have set up a key during Phase-1, you can proceed to
+[Import an existing EOTS key](#422-import-an-existing-eots-key) to import
+your Phase-1 key to the EOTS manager.
 
 ## 2. System Requirements
 
-Recommended specifications for running a Babylon Finality Provider:
+Recommended specifications for running a Finality Provider:
 
 * CPU: 2 vCPUs
 * RAM: 4GB
@@ -80,20 +79,8 @@ These are the minimum specifications for running a finality provider.
 Requirements may vary based on network activity and your operational needs.
 For production environments, you may want to consider using more robust hardware.
 
-**Recovery and Backup**
-At the time of writing, the following assets **must not** be lost and should be
-backed up frequently. Loss will lead to inability to submit transactions to the
-Babylon chain, which will in turn lead to FP jailing and halt BTC Staking reward
-accumulation.
-
-* The `keyring-xx` folder contains your Babylon keyring, used to submit public
-  randomness and finality signatures to Babylon.
-* The `finality-provider.db` contains essential operational data
-  including finality signatures, public randomness proofs, and state information.
-  Loss will prevent voting until recovered.
-
-The ability to recreate the `finality-provider.db` will be offered in the next
-few months.
+For security tips of running a finality provider, please refer to
+[Slashing Protection](./slashing-protection.md), [HMAC Security](./hmac-security.md), and [7. Recovery and Backup](#7-recovery-and-backup).
 
 ## 3. Install Finality Provider Toolset
 
@@ -129,11 +116,11 @@ make install
 
 This command will:
 
-* Build and compile all Go packages
+* Build and compile all Go packages.
 * Install binaries to `$GOPATH/bin`:
   * `eotsd`: EOTS manager daemon
-  * `fpd`: Finality provider daemon
-* Make commands globally accessible from your terminal
+  * `fpd`: Finality provider daemon.
+* Make commands globally accessible from your terminal.
 
 ### 3.3. Verify Installation
 
@@ -152,6 +139,7 @@ depending on your shell.
 
 ```shell
 echo 'export PATH=$HOME/go/bin:$PATH' >> ~/.profile
+source ~/.profile
 ```
 
 ## 4. Setting up the EOTS Daemon
@@ -170,8 +158,10 @@ with the following command:
 eotsd init --home <path>
 ```
 
-If `eotsd.conf` already exists `init` will not succeed, if the operator wishes to
-overwrite the config file they need to use `--force`.
+If the home directly already exists, `init` will not succeed.
+> ⚡ Specifying `--force` to `init` will overwrite `eotsd.conf` with default
+> config values if the home directory already exists.
+> Please backup `eotsd.conf` before you run `init` with `--force`.
 
 Parameters:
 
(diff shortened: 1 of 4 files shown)


docs/finality-provider-operation.md
```diff
@@ -49,24 +49,24 @@ gain an overall understanding of the finality provider.
...
-are required to transition their finality providers to the second phase
+Finality providers that received delegations in Phase-1
+are required to transition their finality providers to Phase-2
 using the same EOTS key that they used and registered with during Phase-1.
 The usage of a different key corresponds to setting up an entirely
 different finality provider which will not inherit the Phase-1 delegations.
-Not transitioning your Phase-1 finality provider prevents your Phase-1 delegations
-from transitioning to the second phase.
+If you don't transition your Phase-1 finality provider to Phase-2, all the
+received Phase-1 delegations cannot transition to Phase-2.
```
vitsalis, line 61:
```suggestion
received Phase-1 delegations won't be able to register on the Babylon Genesis chain.
```

docs/finality-provider-operation.md
```diff
@@ -49,24 +49,24 @@ gain an overall understanding of the finality provider.
...
-Finality providers that received delegations on the first phase of the launch
-are required to transition their finality providers to the second phase
+Finality providers that received delegations in Phase-1
+are required to transition their finality providers to Phase-2
 using the same EOTS key that they used and registered with during Phase-1.
 The usage of a different key corresponds to setting up an entirely
 different finality provider which will not inherit the Phase-1 delegations.
-Not transitioning your Phase-1 finality provider prevents your Phase-1 delegations
-from transitioning to the second phase.
+If you don't transition your Phase-1 finality provider to Phase-2, all the
```
vitsalis, line 60:
```suggestion
If you don't register your Phase-1 finality provider on the Babylon Genesis chain, all the
```

docs/finality-provider-operation.md
```diff
@@ -49,24 +49,24 @@ gain an overall understanding of the finality provider.
...
-Thank you for participating in the first phase of the Babylon launch. This guide
-provides instructions for setting up the full finality provider toolset required
-for your participation in the second phase of the Babylon launch.
+This node is for you if you have participated in Phase-1 of Babylon
+Genesis to help you transition to Phase-2.
 
-Finality providers that received delegations on the first phase of the launch
-are required to transition their finality providers to the second phase
+Finality providers that received delegations in Phase-1
+are required to transition their finality providers to Phase-2
```
vitsalis, line 56:
```suggestion
are required to register their finality providers on the Babylon Genesis chain
```

docs/finality-provider-operation.md
```diff
@@ -80,20 +80,8 @@ These are the minimum specifications for running a finality provider.
...
-* The `keyring-xx` folder contains your Babylon keyring, used to submit public
-  randomness and finality signatures to Babylon.
-* The `finality-provider.db` contains essential operational data
-  including finality signatures, public randomness proofs, and state information.
-  Loss will prevent voting until recovered.
-
-The ability to recreate the `finality-provider.db` will be offered in the next
-few months.
+For security tips of running a finality provider, please refer to
+[Slashing Protection](./slashing-protection.md), [HMAC Security](./hmac-security.md), and [7. Recovery and Backup](#7-recovery-and-backup).
```
vitsalis, line 84: wrap to 80 characters

docs/finality-provider-operation.md
```diff
@@ -221,11 +211,11 @@ Parameters:
 
 ### 4.2. Add an EOTS Key
 
-This section explains the process of setting up the private keys for the
-EOTS manager. Operators *must* create an EOTS key before starting the
+This section explains the process of creating EOTS keys using the EOTS manager.
+Operators *MUST* create an EOTS key before starting the
 EOTS daemon.
```
vitsalis, line 216: Do we have an error if operators don't do that?

docs/finality-provider-operation.md
```diff
@@ -324,28 +313,21 @@ To load the key, use the following command:
 eotsd keys import <name> <path-to-key> --home <path> --keyring-backend test
 ```
 
-#### Option 3: Using a File System Backup
```
vitsalis, line 327: Why is this removed?

docs/finality-provider-operation.md
```diff
@@ -450,11 +431,6 @@ This ensures that the finality provider daemon can promptly submit
 transactions, such as block votes and public randomness submissions that are
 essential for its liveness and earning of rewards.
 
-For the `fpd` keyring, the `test` backend will be exclusively used, and it is
-mandatory that you follow this practice until automated key management becomes
-available. Additionally, we are also exploring options to support different
-withdrawal addresses, so that rewards can go to a separate address.
```
vitsalis, line 456: Why remove this paragraph?

docs/finality-provider-operation.md
```diff
@@ -523,13 +499,13 @@ Configuration parameters explained:
...
-* `RPCAddr`: Your Babylon node's RPC endpoint
+* `Key`: Your Babylon Genesis key name from Step 2
+* `ChainID`: The Babylon Genesis chain ID
+* `RPCAddr`: Your Babylon Genesis node's RPC endpoint
 * `KeyDirectory`: Path to your keyring directory (same as `--home` path)
 
 Please verify the `chain-id` and other network parameters from the official
-[Babylon Networks
+[Babylon Genesis Networks
 repository](https://github.com/babylonlabs-io/networks/tree/main/bbn-test-5/).
```
vitsalis, line 509:
```suggestion
repository](https://github.com/babylonlabs-io/networks).
```

docs/finality-provider-operation.md
```diff
@@ -810,48 +786,38 @@ This command should ask to
...
-refunded by the protocol, we recommend setting up a refunding flow.
-This involves periodically transferring rewards from a beneficiary key to the
-operational key used by the finality provider.
+### 6.3. Set Up Operation Key
 
-#### How to set up the refunding process
+Finality providers consume gas for operations with Babylon Genesis.
+Given that the cost of sending finality signatures is refunded automatically
+after success, the only cost of operating a finality provider is randomness
+commit, which is made periodically with very low gas cost.
```
vitsalis, line 794:
```suggestion
commit, which is made periodically with low gas cost.
```

docs/finality-provider-operation.md
```diff
@@ -810,48 +786,38 @@ This command should ask to
...
 
-1. Ensure you have two keys:
-   * Beneficiary Key: Receives finality provider rewards.
-   * Operational Key: Used by the finality provider daemon to submit transactions.
+Therefore, it is recommended to use a separate key specifically for operations.
+This leaves key used for registration key isolated and only needed for claiming
```
vitsalis, line 797:
```suggestion
This leaves the key used for registration key isolated and only needed for claiming
```

vitsalis: Is this true? I was under the impression that the following holds:

* Key used during registration **must** be the same as the one used for operations.
* However, the registration key can be used to set a separate withdraw address to a more secure keyring than the one used for ops.

From the text it looks like the register key has to be the secure one, while there's a gas-payment key. How are the two keys associated, with the Babylon node knowing that operations (e.g. randomness commits) are coming from the owner of the register key?

docs/finality-provider-operation.md
```diff
@@ -810,48 +786,38 @@ This command should ask to
...
+rewards. Given that the operation key is totally replaceable, it can hold only
```
vitsalis, line 798: How is it replaceable?

docs/finality-provider-operation.md
```diff
@@ -810,48 +786,38 @@ This command should ask to
...
+3. Set the operation key name (`Key`) and the keyring directory (`KeyDirectory`)
+   under `[babylon]` in `fpd.conf`.
 
-Only maintain the minimum balance required for finality provider operations in
-the operational key as this is a hot key. Excess funds should be kept safely
-in the benefiary address.
+4. Restart the finality provider daemon.
 
 > 💡 **Tip**: Committing randomness has a constant cost of `0.000130BBN` per
 > commit. Therefore, reserving `5-10bbn` for operations should be enough for a
```
vitsalis, line 815: The denom is BABY

vitsalis (approved):
lgtm, some minor comments and a question about the operational and withdraw keyrings

maiquanghiep:
I think we can add the `--node` flag to the command so that operator would know they have to put the RPC endpoint
https://github.com/babylonlabs-io/finality-provider/blob/2a659125f2390af08fa328a79236dd974a07a136/docs/finality-provider-operation.md?plain=1#L720

maiquanghiep:
Should we move the `Set Withdraw Address` step before `Withdraw Rewards` to make it clear that the beneficiary address can be set before withdrawing?
https://github.com/babylonlabs-io/finality-provider/blob/main/docs/finality-provider-operation.md?plain=1#L751

docs/finality-provider-operation.md
gitferry: Nope. We probably can remove Line 215

docs/finality-provider-operation.md
gitferry: Any key can be the operation key as long as it's funded. It's not bonded to the fp. Will rephrase as following:

```
The operation key is totally replaceable in the sense that it is not
tied to a finality provider, and it can be replaced by any key that is properly
funded.
```

docs/finality-provider-operation.md
```diff
@@ -49,24 +49,24 @@ gain an overall understanding of the finality provider.
 
 ## 1. A note about Phase-1 Finality Providers
 
-Thank you for participating in the first phase of the Babylon launch. This guide
-provides instructions for setting up the full finality provider toolset required
-for your participation in the second phase of the Babylon launch.
+This node is for you if you have participated in Phase-1 of Babylon
```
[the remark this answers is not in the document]
vitsalis: Prefer the previous one tbh, but no strong opinions here

docs/finality-provider-operation.md
vitsalis: If they must do that, shouldn't we have an error?

docs/finality-provider-operation.md
gitferry: Will reply offline.

gitferry:
> Should we move the Set Withdraw Address step before Withdraw Rewards to make it clear that the beneficiary address can be set before withdrawing?
https://github.com/babylonlabs-io/finality-provider/blob/main/docs/finality-provider-operation.md?plain=1#L751

I was worried that this will make people think `set-withdraw-addr` is a pre-step for `withdraw-rewards`. Will add a reference in `withdraw-rewards` to `set-withdraw-addr` if people want to change the benefiary

docs/finality-provider-operation.md
gitferry: Maybe we can add a check during eotsd initiation. This check should examine the db record for key creation and the program should exist if it find none. There's no point of starting eotsd anyway if no key is created

PR #466, opened 2025-05-08
feat: adds hot backup for eotsd

Lazar955:
adds 🔥 backup for eots.db, enable operators to backup while the eotsd is running.

Uses the idiomatic underlying `copy` mechanism of [bbolt](https://github.com/etcd-io/bbolt/blob/c3550d92a41267a9a7dcfde6db3c05422e57e94d/tx.go#L382).

Test results when running the Copy (MBP M3):

| Database Entries | Avg. Backup Time | Avg. Backup Size |
| ---------------- | ---------------- | ---------------- |
| 1000             | 443.694333µs     | 0.56 MB          |
| 50000            | 10.361083ms      | 17.17 MB         |
| 150000           | 15.753723ms      | 38.89 MB         |
| 1000000          | 78.167639ms      | 208.66 MB        |


RafilxTenfen:
Wouldn't it be better if the DB wasn't busy all the time, like instead of having a backup call, it only keeps the db file locked when it needs to use it and frees it right after. Then, if someone needs to copy the files, they can do normal `$~ cp eotsd.db eotsd-bckp.db`

Lazar955:
> Wouldn't it be better if the DB wasn't busy all the time, like instead of having a backup call, it only keeps the db file locked when it needs to use it and frees it right after. Then, if someone needs to copy the files, they can do normal `$~ cp eotsd.db eotsd-bckp.db`

I think in that case we would have to fork bbolt? I don't think removing/adding lock is idiomatic, there's a reason why they expose the `Copy` interface

eotsmanager/cmd/eotsd/daemon/backup.go
```diff
@@ -0,0 +1,65 @@
...
+
+	return cmd
+}
+
+func backup(cmd *cobra.Command, _ []string) error {
+	f := cmd.Flags()
+
+	dbPath, err := f.GetString(flagDBPath)
+	if err != nil {
+		return fmt.Errorf("failed to get db-path flag: %w", err)
```
gitferry, line 40:
```suggestion
		return fmt.Errorf("failed to get %s flag: %w", flagDBPath, err)
```

eotsmanager/cmd/eotsd/daemon/backup.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	backupDir, err := f.GetString(flagBackupDir)
+	if err != nil {
+		return fmt.Errorf("failed to get backup-dir flag: %w", err)
```
gitferry, line 45:
```suggestion
		return fmt.Errorf("failed to get %s flag: %w", flagBackupDir, err)
```

eotsmanager/store/eotsstore.go
```diff
@@ -292,3 +294,37 @@ func ExtractHeightFromKey(key []byte) (uint64, error) {
...
+	if dbPath == "" {
+		return fmt.Errorf("database path must be provided")
+	}
+
+	if backupDir == "" {
+		return fmt.Errorf("backup dir path must be provided")
+	}
+
+	// Create backup filename with timestamp
+	backupName := "eots.db"
```
gitferry, line 309: Would it be better to let user specify the backup name? Maybe they want to add timestamp or something

eotsmanager/cmd/eotsd/daemon/flags.go
```diff
@@ -17,5 +17,7 @@ const (
 	flagHDPath            = "hd-path"
 	flagIndex             = "index"
 	flagRecover           = "recover"
-	flagMnemonicSrc       = "source"
+	flagMnemonicSrc = "source"
+	flagDBPath      = "db-path"
+	flagBackupDir   = "backup-dir"
```
gitferry, line 22: nit: not alligned in style

eotsmanager/eotsmanager.go
```diff
@@ -31,5 +31,9 @@ type EOTSManager interface {
 	// or passPhrase is incorrect
 	SignSchnorrSig(uid []byte, msg []byte) (*schnorr.Signature, error)
 
+	// Backup performs a hot backup of the database using a read-only transaction, eotsd can be running
+	// when this function is called, but writing to the db is blocked until the backup is done
+	Backup(dbPath string, backupDir string) error
```
gitferry, line 36: Do we know how long the db will be blocked? I was worried that if the db file is super big, it will affect the daemon running. Can we have a limit of the backup? For example, only backup the latest voting history records

gitferry (commented):
I'm a bit worried about the time of backup as it seems will block the signing process. Maybe we can do:
1. benchmark the backup
2. explore ways of partial backup (only backup latest records)

Lazar955:
> I'm a bit worried about the time of backup as it seems will block the signing process. Maybe we can do:
> 
> 1. benchmark the backup
> 2. explore ways of partial backup (only backup latest records)

Yep, it will block the signing process. But the fuzz test demonstrates that even while backup is happening, the original DB contains all the records expected during concurrency writes, and that the backup contains up to the start of the backup.

1. Benchmark is a good idea, will add it so we understand how long it takes to copy a big db.
2. I don't think there's a way for a partial backup, mostly because the keys are not ordered (not explicitly through interface). But that being said, could we have pruning for voting records?

eotsmanager/eotsmanager.go
Lazar955: Added benchmark, seems that for 1m records ~200mb, copy operations takes ~200ms

RafilxTenfen (approved):
LGTM, code for hot backup makes sense to me

In my opinion in the future we should still look into only keep the db file locked while using it

gitferry (approved):
LGTM! Great work!

gitferry:
> In my opinion in the future we should still look into only keep the db file locked while using it

@RafilxTenfen Good point. Ideally, we should not lock the read permission. @Lazar955 Could you create an issue for this?

eotsmanager/store/eotsstore.go
```diff
@@ -292,3 +294,37 @@ func ExtractHeightFromKey(key []byte) (uint64, error) {
...
+	backupPath := filepath.Join(backupDir, backupName)
+
+	// Create the directory if it doesn't exist
+	if err := os.MkdirAll(backupDir, 0750); err != nil {
```
KonradStaniec, line 313: Wonder whether we do not need to add some lock over whole `BackupDB`  function to make sure it is not called concurrently ? 

Is `os.MkdirAll` will only create dir if id does on exists so two concurrent calls are probably fine, I am more concerned about `os.Create(backupPath)` as this may open file from two threads?

eotsmanager/store/eotsstore.go
```diff
@@ -292,3 +294,37 @@ func ExtractHeightFromKey(key []byte) (uint64, error) {
...
+
+	// Open the backup file
+	// #nosec G304 -- backupPath is provided by operators
+	backupFile, err := os.Create(backupPath)
+	if err != nil {
+		return fmt.Errorf("failed to create backup file: %w", err)
+	}
+	defer backupFile.Close()
+
+	if err := s.db.Copy(backupFile); err != nil {
```
KonradStaniec, line 325: To understand operations when this runs:
- backcup is initiated
- writes are blocked
- so when FP asks EOTSd to sign, eotsd won't return until this operation finishes as any new signature generates write to anti slashing db
- after backup finishes, we write the requested sig to db (but not to the backup)
- now we return sig to the caller

So during whole backup procedure, at most one signature can be written to the orignal DB but not to the backup.

Lazar955: Exactely

eotsmanager/store/eotsstore.go
Lazar955: I think you are right here. Was looking over the `copy` function on bbolt, they don't lock the backup file for the duration of it, so we should make sure we lock. 
That being said this is executed from the CLI so concurrency might not be a concern, but it's still valid. I'll add the fix. Thanks

Lazar955: Added mutex in [886d1cf](https://github.com/babylonlabs-io/finality-provider/pull/466/commits/886d1cfb1764c39191865567f25b3610f57ce4d4)

PR #469, opened 2025-05-09
chore: adds e2e for unsafe commit pub rand

Lazar955:
To commit 100k pub rand takes: `8m11s`


finality-provider/service/fp_instance.go
```diff
@@ -151,6 +151,12 @@ func (fp *FinalityProviderInstance) Stop() error {
 	close(fp.quit)
 	fp.wg.Wait()
 
+	if closeDB {
```
Lazar955, line 154: not ideal, but wrangling start of the db for tests is a bit annoying

gitferry:
Do we still want to merge this pr?

Lazar955:
> Do we still want to merge this pr?

Having an additional test doesn't hurt, so I'd say yes

finality-provider/service/fp_instance.go
gitferry: Maybe we have have a separate stop for test, i.e., `StopWithDBClosed`

Lazar955: yeah good idea!

Lazar955: adde in [a4d85d4](https://github.com/babylonlabs-io/finality-provider/pull/469/commits/a4d85d46d01b391a13f945a83208ee5865eb0cab)

gitferry (approved):
Great work!

PR #475, opened 2025-05-28
chore: bump bbn to v2.0.0-rc.0

Lazar955:


Lazar955:
> should we bump main to v4 as well?

I think it make sense, I'll open a PR

PR #482, opened 2025-06-04
chore: audit comments

Lazar955:


eotsmanager/eotsmanager.go
```diff
@@ -28,12 +28,11 @@ type EOTSManager interface {
...
 
 	// SignSchnorrSig signs a Schnorr signature using the private key of the finality provider
-	// It fails if the finality provider does not exist or the message size is not 32 bytes
-	// or passPhrase is incorrect.
+	// It fails if the finality provider does not exist or the message size is not 32 bytes.
 	// uid represents EOTS key
 	SignSchnorrSig(uid []byte, msg []byte) (*schnorr.Signature, error)
 
-	// Unlock makes the private key for the given EOTS key (uid) accessible in memory using the provided passphrase.
+	// Unlock makes the private key for the given EOTS key (uid) accessible in memory using the provided keyring passphrase.
```
RafilxTenfen, line 35:
```suggestion
	// Unlock makes the private key for the given EOTS key (uid) accessible in memory using the provided keyring key password.
```

Should we rename to password? I believe it is called password by the keyring file in cosmos https://github.com/cosmos/keyring/blob/master/file.go#L31

eotsmanager/localmanager.go
```diff
@@ -121,8 +123,10 @@ func (lm *LocalEOTSManager) CreateKeyWithMnemonic(name, mnemonic, passphrase str
 		return nil, err
 	}
 
+	// when the first key is created for the `file` keyring backend, it will prompt for a passphrase twice
```
RafilxTenfen, line 126: They call passwords in the variables, but passphrase in in the prompt https://github.com/cosmos/keyring/blob/master/file.go#L61

:exploding_head:

RafilxTenfen (approved):
LGTM

Issue #486, opened 2025-06-05
Update docs to reflect Unlock command


Lazar955:
- Update docs for unlock command
- Update docs that we support file keyring also
- Update docs how to migrate test -> file keyring

PR #487, opened 2025-06-06
chore: unlock cmd hmac and password from input

Lazar955:
- Provide the user with the option to use HMAC.
- Use a more secure way to read the password for keyring unlock.

### DIFF
--- a/eotsmanager/cmd/eotsd/daemon/flags.go
+++ b/eotsmanager/cmd/eotsd/daemon/flags.go
@@ -18,7 +18,8 @@ const (
 	flagIndex             = "index"
 	flagRecover           = "recover"
 	flagMnemonicSrc       = "source"
-	passphraseFlag        = "passphrase"
-	flagDBPath      = "db-path"
-	flagBackupDir   = "backup-dir"
+	flagPassphrase        = "passphrase"
+	flagDBPath            = "db-path"
+	flagBackupDir         = "backup-dir"
+	flagHMAC              = "hmac"
 )
--- a/eotsmanager/cmd/eotsd/daemon/unlock.go
+++ b/eotsmanager/cmd/eotsd/daemon/unlock.go
@@ -5,7 +5,18 @@ import (
 	"fmt"
 	eotsclient "github.com/babylonlabs-io/finality-provider/eotsmanager/client"
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
+	"os"
 )
+
+var UnlockCmdPasswordReader = passwordReader
+
+func passwordReader(cmd *cobra.Command) (string, error) {
+	cmd.Print("Enter password to unlock keyring: ")
+	passphrase, err := term.ReadPassword(int(os.Stdin.Fd()))
+
+	return string(passphrase), err
+}
 
 func NewUnlockKeyringCmd() *cobra.Command {
 	cmd := &cobra.Command{
@@ -17,13 +28,10 @@ func NewUnlockKeyringCmd() *cobra.Command {
 	f := cmd.Flags()
 
 	f.String(eotsPkFlag, "", "EOTS public key of the finality-provider")
-	f.String(passphraseFlag, "", "The keyring passphrase")
 	f.String(rpcClientFlag, "", "The RPC address of a running eotsd")
+	f.String(flagHMAC, "", "When using HMAC either pass here as flag or set env variable HMAC_KEY.")
 
 	if err := cmd.MarkFlagRequired(eotsPkFlag); err != nil {
-		panic(err)
-	}
-	if err := cmd.MarkFlagRequired(passphraseFlag); err != nil {
 		panic(err)
 	}
 	if err := cmd.MarkFlagRequired(rpcClientFlag); err != nil {
@@ -38,22 +46,32 @@ func unlockKeyring(cmd *cobra.Command, _ []string) error {
 
 	eotsFpPubKeyStr, err := f.GetString(eotsPkFlag)
 	if err != nil {
-		return fmt.Errorf("failed to get eots pk flag: %w", err)
-	}
-
-	passphrase, err := f.GetString(passphraseFlag)
-	if err != nil {
-		return fmt.Errorf("failed to get chain-id flag: %w", err)
+		return fmt.Errorf("failed to get %s flag: %w", eotsPkFlag, err)
 	}
 
 	rpcListener, err := cmd.Flags().GetString(rpcClientFlag)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get %s flag: %w", rpcClientFlag, err)
 	}
 
-	eotsdClient, err := eotsclient.NewEOTSManagerGRpcClient(rpcListener, "")
+	hmac, err := cmd.Flags().GetString(flagHMAC)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get %s flag: %w", flagHMAC, err)
+	}
+
+	envHmac, exists := os.LookupEnv("HMAC_KEY")
+	if hmac == "" && exists {
+		hmac = envHmac
+	}
+
+	passphrase, err := UnlockCmdPasswordReader(cmd)
+	if err != nil {
+		return fmt.Errorf("failed to read passphrase: %w", err)
+	}
+
+	eotsdClient, err := eotsclient.NewEOTSManagerGRpcClient(rpcListener, hmac)
+	if err != nil {
+		return fmt.Errorf("failed to create eotsd client: %w", err)
 	}
 
 	eotsFpPub, err := hex.DecodeString(eotsFpPubKeyStr)
--- a/itest/babylon/babylon_e2e_test.go
+++ b/itest/babylon/babylon_e2e_test.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/spf13/cobra"
 	"log"
 	"math/rand"
 	"os"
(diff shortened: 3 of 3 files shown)


eotsmanager/cmd/eotsd/daemon/unlock.go
```diff
@@ -5,8 +5,19 @@ import (
 	"fmt"
 	eotsclient "github.com/babylonlabs-io/finality-provider/eotsmanager/client"
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
```
vitsalis, line 8: Is this the same package we use on the babylon node for bls keys? If not, why did we choose a different one?

eotsmanager/cmd/eotsd/daemon/unlock.go
```diff
@@ -17,15 +28,12 @@ func NewUnlockKeyringCmd() *cobra.Command {
 	f := cmd.Flags()
 
 	f.String(eotsPkFlag, "", "EOTS public key of the finality-provider")
-	f.String(passphraseFlag, "", "The keyring passphrase")
 	f.String(rpcClientFlag, "", "The RPC address of a running eotsd")
+	f.String(flagHMAC, "", "When using HMAC either pass here as flag or set env variable HMAC_KEY.")
```
vitsalis, line 32: Based on the [docs](https://github.com/babylonlabs-io/finality-provider/blob/main/docs/hmac-security.md#hmac-security-for-finality-provider), we already have an HMAC key as part of the configuration for interaction between EOTS manager and finality provider.

Is this the same hmac key? If not, should this hmac key also be specified in the configuration?

eotsmanager/cmd/eotsd/daemon/unlock.go
```diff
@@ -38,22 +46,32 @@ func unlockKeyring(cmd *cobra.Command, _ []string) error {
...
-	rpcListener, err := cmd.Flags().GetString(rpcClientFlag)
+	hmac, err := cmd.Flags().GetString(flagHMAC)
+	if err != nil {
+		return fmt.Errorf("failed to get %s flag: %w", flagHMAC, err)
+	}
+
+	envHmac, exists := os.LookupEnv("HMAC_KEY")
+	if hmac == "" && exists {
+		hmac = envHmac
+	}
```
vitsalis, line 65: The functionality for retrieving the hmac key should be in a separate function?

eotsmanager/cmd/eotsd/daemon/unlock.go
vitsalis: Also, this is not a proper flag description. It provides instructions instead of describing what the flag does. It should describe what the value is used for and what it defaults to (i.e., env)

eotsmanager/cmd/eotsd/daemon/unlock.go
```diff
@@ -17,15 +28,12 @@ func NewUnlockKeyringCmd() *cobra.Command {
 	f := cmd.Flags()
 
 	f.String(eotsPkFlag, "", "EOTS public key of the finality-provider")
-	f.String(passphraseFlag, "", "The keyring passphrase")
```
vitsalis, line 20: Should the user have the option to provide the password in other ways?

eotsmanager/cmd/eotsd/daemon/unlock.go
Lazar955: `term` doesn't echo input in the terminal (input not visible). On babylon node we used `fmt.Scanln`. This approach is better

eotsmanager/cmd/eotsd/daemon/unlock.go
Lazar955: Might make sense to support env var as well?

eotsmanager/cmd/eotsd/daemon/unlock.go
Lazar955: I dont see a reason why this should be a different key. Will update the str

Lazar955: I created an issue to update documentation for the Unlock command, let's push this PR forward

Lazar955: See https://github.com/babylonlabs-io/finality-provider/issues/486

eotsmanager/cmd/eotsd/daemon/flags.go
```diff
@@ -18,7 +18,7 @@ const (
 	flagIndex             = "index"
 	flagRecover           = "recover"
 	flagMnemonicSrc       = "source"
-	passphraseFlag        = "passphrase"
-	flagDBPath      = "db-path"
-	flagBackupDir   = "backup-dir"
+	flagDBPath            = "db-path"
+	flagBackupDir         = "backup-dir"
+	flagHMAC              = "hmac"
```
gitferry, line 23: hmac is also a secret and could be exposed if user specifies it via grpc?

eotsmanager/cmd/eotsd/daemon/unlock.go
```diff
@@ -5,27 +5,37 @@ import (
...
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
+	"os"
 )
 
+var UnlockCmdPasswordReader = passwordReader
+
+func passwordReader(cmd *cobra.Command) (string, error) {
+	cmd.Print("Enter password to unlock keyring: ")
+	passphrase, err := term.ReadPassword(int(os.Stdin.Fd()))
```
gitferry, line 16: given the unlock cmd can be executed remotely, will the remote operator get the prompt for password?

eotsmanager/cmd/eotsd/daemon/unlock.go
```diff
@@ -5,27 +5,37 @@ import (
...
+}
+
 func NewUnlockKeyringCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "unlock",
-		Short: "Unlocks the file based keyring to load the EOTS private key in memory for signing",
-		RunE:  unlockKeyring,
+		Short: `Unlocks the "file" based keyring to load the EOTS private key in memory for signing`,
+		Long: `Unlocks the "file" based keyring. Keyring password can be provided either through environment variable 
+"EOTSD_KEYRING_PASSWORD" or through input when executing the command. If the "EOTSD_KEYRING_PASSWORD" the command doesn't prompt for password.'`,
```
gitferry, line 26:
```suggestion
"EOTSD_KEYRING_PASSWORD" or through input when executing the command. If the "EOTSD_KEYRING_PASSWORD" is filled, the command doesn't prompt for password.'`,
```

gitferry:
Maybe not for this pr, but I suggest we update the operation doc to add the instructions of using the `unlock` cmd

eotsmanager/cmd/eotsd/daemon/unlock.go
vitsalis: Given that we are updating the unlock command in this PR, shall we also update the documentation for it? For the rest of the docs it makes sense. Otherwise, the new version will have outdated docs.

vitsalis: What about the HMAC key in the config? Am I missing something?

eotsmanager/cmd/eotsd/daemon/unlock.go
Lazar955: yes

docs/finality-provider-operation.md
```diff
@@ -373,6 +376,27 @@ this value by specifying a custom address with the `--rpc-listener` flag.
...
+
+```shell
+export EOTSD_KEYRING_PASSWORD=<your-password>
+```
+
+If you have HMAC security enabled, you can also specify the HMAC key either with:
+- `HMAC_KEY` environment variable
+- Providing the `--home` path to the eotsd home directory which contains the config file.
+
 >**🔒 Security Tip**:
```
vitsalis, line 400: The security tip seems to be for the 5.3.1. section, but it's a general one, I suggest we move it before the section as before.

docs/finality-provider-operation.md
```diff
@@ -373,6 +376,27 @@ this value by specifying a custom address with the `--rpc-listener` flag.
 EOTS Manager Daemon is fully active!
 ```
 
+#### 5.3.1. Unlock file-based keyring
+
+⚠️ If you are using a `file` based keyring-backend, you need to unlock the keyring by executing the following command:
```
vitsalis, line 381: When does someone need to unlock their keyring (e.g., after starting the daemon) and why (e.g., to be able to sign requests?)? If they don't do so, what happens (e.g., the program will error)?

After starting the daemon? And also in order to

docs/finality-provider-operation.md
```diff
@@ -373,6 +376,27 @@ this value by specifying a custom address with the `--rpc-listener` flag.
...
+
+```shell
+eotsd unlock --eots-pk <eots-pk> --rpc-client <eotsd-address>
+```
+
+You will be prompted to enter the password for the keyring. After which the signing operations will be available and eotsd can
+run uninterrupted.
+
+Alternatively to inputting the password, you can specify the password with the environment variable `EOTS_KEYRING_PASSWORD`:
```
vitsalis, line 390:
```suggestion
An alternative to providing the password as an input is specifying the password with the environment variable `EOTS_KEYRING_PASSWORD`:
```

docs/finality-provider-operation.md
```diff
@@ -373,6 +376,27 @@ this value by specifying a custom address with the `--rpc-listener` flag.
...
+
+```shell
+export EOTSD_KEYRING_PASSWORD=<your-password>
+```
+
+If you have HMAC security enabled, you can also specify the HMAC key either with:
+- `HMAC_KEY` environment variable
+- Providing the `--home` path to the eotsd home directory which contains the config file.
```
vitsalis, line 398:
```suggestion
* Providing the `--home` path to the eotsd home directory which contains the config file
   containing your HMAC password.
```

docs/finality-provider-operation.md
```diff
@@ -373,6 +376,27 @@ this value by specifying a custom address with the `--rpc-listener` flag.
...
+run uninterrupted.
+
+Alternatively to inputting the password, you can specify the password with the environment variable `EOTS_KEYRING_PASSWORD`:
+
+```shell
+export EOTSD_KEYRING_PASSWORD=<your-password>
+```
+
+If you have HMAC security enabled, you can also specify the HMAC key either with:
+- `HMAC_KEY` environment variable
```
vitsalis, line 397:
```suggestion
* `HMAC_KEY` environment variable
```

for consistency in the doc

vitsalis (approved):
small comments

PR #493, opened 2025-06-09
chore: key migration for test keyring

Lazar955:
WIP:

adds flag `--migrate=true` to skip two unnecessary prompts when saving to DB and log

### DIFF
--- a/eotsmanager/cmd/eotsd/daemon/flags.go
+++ b/eotsmanager/cmd/eotsd/daemon/flags.go
@@ -20,4 +20,5 @@ const (
 	flagMnemonicSrc       = "source"
 	flagDBPath            = "db-path"
 	flagBackupDir         = "backup-dir"
+	flagMigrate           = "migrate"
 )
--- a/eotsmanager/cmd/eotsd/daemon/keys.go
+++ b/eotsmanager/cmd/eotsd/daemon/keys.go
@@ -67,6 +67,8 @@ func NewKeysCmd() *cobra.Command {
 	}
 
 	addCmd.Flags().String(rpcClientFlag, "", "The RPC address of a running eotsd to connect and save new key")
+	addCmd.Flags().Bool(flagMigrate, false, "Used during key migration to skip printing and saving "+
+		"the key name mapping, avoiding redundant keyring unlock prompts.")
 
 	// Override the original RunE function to run almost the same as
 	// the sdk, but it allows empty hd path and allow to save the key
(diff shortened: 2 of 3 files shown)


RafilxTenfen:
> adds flag --migrate=true to skip saving to the eotsd DB

Why do we want to skip saving ?

eotsmanager/cmd/eotsd/daemon/keys.go
```diff
@@ -60,38 +60,45 @@ func NewKeysCmd() *cobra.Command {
...
 
-			cmd.Printf("Key Name: %s\nEOTS PK: %s\n", args[0], eotsPk.MarshalHex())
-
-			return nil
+			return printFromKey(cmd, clientCtx, args[0], eotsPk)
 		}
 	}
 
 	addCmd.Flags().String(rpcClientFlag, "", "The RPC address of a running eotsd to connect and save new key")
+	addCmd.Flags().Bool(flagMigrate, false, "Used during key migration to skip saving the key into eotsd.db")
```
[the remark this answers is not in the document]
RafilxTenfen: Could we check the error type or something instead of having another flag just for this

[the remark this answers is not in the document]
Lazar955: Yep, valid point, removed the flag

RafilxTenfen (approved):
it looks good to me, I can create another key with the same name 
I am not sure whether it is expected that I can create a new key with the same name without checking if the eots pk corresponding to that name matches the old one or maybe propmt ask the user when there is an `ErrDuplicateEOTSKeyName` 

So I am approving, but might be good to ask for @KonradStaniec to take a look

eotsmanager/cmd/eotsd/daemon/flags.go
```diff
@@ -20,4 +20,5 @@ const (
 	flagMnemonicSrc       = "source"
 	flagDBPath            = "db-path"
 	flagBackupDir         = "backup-dir"
+	flagMigrate           = "migrate"
```
gitferry, line 23: seems we don't need this flag

eotsmanager/cmd/eotsd/daemon/keys.go
```diff
@@ -134,22 +141,22 @@ func saveKeyNameMapping(cmd *cobra.Command, keyName string) (*types.BIP340PubKey
...
 		if err != nil {
 			return nil, fmt.Errorf("failed to get public key for key %s: %w", keyName, err)
 		}
 
-		if err := client.SaveEOTSKeyName(eotsPk.MustToBTCPK(), keyName); err != nil {
+		if err := eotsClient.SaveEOTSKeyName(eotsPk.MustToBTCPK(), keyName); err != nil {
+			// ignore the err, keyring will handle it
+			if errors.Is(err, store.ErrDuplicateEOTSKeyName) {
+				return eotsPk, nil
+			}
```
gitferry, line 158: We probably need to check if the key is the same as the key with the same key name. If not, we probably should reject

eotsmanager/cmd/eotsd/daemon/keys.go
```diff
@@ -176,13 +183,18 @@ func saveKeyNameMapping(cmd *cobra.Command, keyName string) (*types.BIP340PubKey
...
 
 	// Save the public key to key name mapping
 	if err := eotsManager.SaveEOTSKeyName(eotsPk.MustToBTCPK(), keyName); err != nil {
+		// ignore the err, keyring will handle it
+		if errors.Is(err, store.ErrDuplicateEOTSKeyName) {
+			return eotsPk, nil
+		}
+
 		return nil, fmt.Errorf("failed to save key name mapping: %w", err)
 	}
```
gitferry, line 199: Seems duplicate with Line 154. Let's wrap it

eotsmanager/cmd/eotsd/daemon/keys.go
```diff
@@ -176,13 +183,23 @@ func saveKeyNameMapping(cmd *cobra.Command, keyName string) (*types.BIP340PubKey
...
 
 	// Save the public key to key name mapping
 	if err := eotsManager.SaveEOTSKeyName(eotsPk.MustToBTCPK(), keyName); err != nil {
+		// ignore the err, keyring will handle it
+		if errors.Is(err, store.ErrDuplicateEOTSKeyName) {
+			return eotsPk, nil
+		}
+
+		if errors.Is(err, store.ErrDuplicateEOTSKeyRecord) {
```
Lazar955, line 198: @RafilxTenfen @gitferry 

Either we just warn the user here. Or we delete the key by calling clientCtx.Keyring.Delete(....)

Lazar955: sg, I'll fix unit tests and merge. Thanks for reviews everyone!

PR #494, opened 2025-06-10
F/consumer spec

maurolacy:
Spec for the finality provider Consumer adapter.


RafilxTenfen (approved):
LGTM

Lazar955 (approved):
nice work, should we create an issue to divide the ConsumerController like it's described here

Issue #503, opened 2025-07-06
rollup: remove dependency to finality gadget in rollup BSN FP


0xAleksaOpacic:
## Description:

Currently, the OP client controller depends on the off-chain Finality Gadget to function, specifically for calling QueryActivatedHeight which internally uses QueryBtcStakingActivatedTimestamp:
https://github.com/babylonlabs-io/finality-provider/blob/main/clientcontroller/opstackl2/consumer.go#L449-L472

🔴 Problem:
This introduces an unnecessary dependency, as every BSN may implement or choose not to implement their own version of the Finality Gadget. Finality Providers (FPs) should not rely on it or have any references to it.

⚠️ Some external teams have worked around this by manually setting the activated height via a flag, but this is not ideal.

✅ Action:
We need to define a better way for the OP client controller to handle activation height without relying on the Finality Gadget.

PR #504, opened 2025-07-06
opstackl2: temporarily disable Finality Gadget dependency

0xAleksaOpacic:
## Description 
This PR is a quick fix to unblock testing, as we cannot rely on the Finality Gadget being present in all setups.

Note: This is a temporary change and should not be merged. The proper fix is tracked in [#503](https://github.com/babylonlabs-io/finality-provider/issues/503).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,9 @@ build/
 dist/
 
 *.swp
+
+# Local directories for testing Finality Providers and EOTS
+anvilFpHome/
+anvilEotsHome/
+babylonFpHome/
+babylonEotsHome/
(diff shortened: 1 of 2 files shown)


SebastianElvis:
> I still don't understand concept of activation hight. Can you write setnance or two just to make sure we are aligned before we make decision

In Babylon chain, the activation height is the height since which the chain starts receiving BTC staking finality, i.e., submitting finality signatures are allowed and the chain will tally these finality signatures.

In the finality provider implementation, it will only submit finality signatures since this height.

The actviation height in the BSN context, however, depends on the implementation of the finality gadget. A proper design would be to defer the responsibility to set the right actviation height to individual finality providers. Wdyt?

SebastianElvis:
Fixed in https://github.com/babylonlabs-io/finality-provider/pull/522 by hardcoding 0 as activation height. Will work on a proper fix in subsequent PR

PR #539, opened 2025-07-18
api: replace `QueryLatestBlockHeight` with `QueryLatestBlock`

### LINKED ISSUE #332: controller: replace `QueryLatestBlockHeight` with `QueryLatestBlock`
In most cases `QueryLatestBlockHeight` is implemented through querying the block and returning its height. This sometimes leads to one extra query. It would be better to return the block directly, like in v1.

### LINKED ISSUE #344: avoid using `zap.NewNop()` logger
non-op logger is extremely hard to debug issue. it caused us a few hours to find out why our op e2e test failed with no error

we are now changing it to 

```
	loggerConfig := zap.NewDevelopmentConfig()
	loggerConfig.Level = zap.NewAtomicLevelAt(zap.ErrorLevel)
	logger, err := loggerConfig.Build()
```

and now able to see the error

```
2024-06-29T18:05:59.836743Z	fatal	terminating the finality-provider instance due to critical error	{"pk": "97276691ab7632441c0d1134d50757387384e382bb2e6e136ec0cfa434e903b7", "error": "reached max failed cycles with err: failed to get inclusion proof of public randomness: public randomness proof not found"}
FAIL	github.com/babylonchain/finality-provider/itest/opstackl2	58.770s
FAIL
make: *** [test-e2e-op] Error 1
```

we fixed it in our test but by quickly searching the codebase, we found many other places are using it.

SebastianElvis:
Resolves #332 
Also resolves #344 when modifying tests


Lazar955 (approved):
lgtm, minor comments

PR #542, opened 2025-07-21
chore: abstract finality submitter

Lazar955:
- Abstracts away finality submission service logic

### DIFF
--- a/finality-provider/service/fp_instance.go
+++ b/finality-provider/service/fp_instance.go
@@ -112,8 +112,9 @@ func newFinalityProviderInstanceFromStore(
 	btcPk := bbntypes.NewBIP340PubKeyFromBTCPK(sfp.BtcPk)
 	fpState := NewFpState(sfp, s, logger, metrics)
 
-	rndCommitter.SetBtcPk(btcPk)
-	rndCommitter.SetChainID([]byte(sfp.ChainID))
+	if err := rndCommitter.Init(btcPk, []byte(sfp.ChainID)); err != nil {
+		return nil, fmt.Errorf("failed to initialize randomness committer: %w", err)
+	}
 	if err := finalitySubmitter.InitState(fpState); err != nil {
 		return nil, fmt.Errorf("failed to initialize finality submitter state: %w", err)
 	}
--- a/finality-provider/service/rand_committer.go
+++ b/finality-provider/service/rand_committer.go
@@ -72,12 +72,22 @@ func NewDefaultRandomnessCommitter(
 	}
 }
 
-func (rc *DefaultRandomnessCommitter) SetBtcPk(btcPk *bbntypes.BIP340PubKey) {
-	rc.btcPk = btcPk
-}
+func (rc *DefaultRandomnessCommitter) Init(btcPk *bbntypes.BIP340PubKey, chainID []byte) error {
+	if btcPk == nil {
+		return fmt.Errorf("btcPk cannot be nil")
+	}
+	if len(chainID) == 0 {
+		return fmt.Errorf("chainID cannot be empty")
+	}
 
-func (rc *DefaultRandomnessCommitter) SetChainID(chainID []byte) {
+	if rc.btcPk != nil && rc.cfg.ChainID != nil {
+		return fmt.Errorf("randomness committer is already initialized with btcPk and chainID")
+	}
+
+	rc.btcPk = btcPk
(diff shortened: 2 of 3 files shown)


finality-provider/service/app.go
```diff
@@ -34,16 +34,17 @@ type FinalityProviderApp struct {
...
+	consumerCon       ccapi.ConsumerController
+	kr                keyring.Keyring
+	fps               *store.FinalityProviderStore
+	pubRandStore      *store.PubRandProofStore
+	config            *fpcfg.Config
+	logger            *zap.Logger
+	poller            types.BlockPoller[types.BlockDescription]
+	rndCommitter      types.RandomnessCommitter
+	heightDeterminer  types.HeightDeterminer
+	finalitySubmitter types.FinalitySignatureSubmitter
```
SebastianElvis, line 47: Since all finality related operations are done within `fpIns`, can we remove some fields at L37-47 from `FinalityProviderApp`?

For example, can we remove L44-47 and invoke `fpIns` for all finality related operations?

finality-provider/service/fp_store_adapter.go
```diff
@@ -12,19 +14,27 @@ import (
 	"github.com/babylonlabs-io/finality-provider/finality-provider/store"
 )
 
+var _ types.FinalityProviderState = (*FpState)(nil)
```
SebastianElvis, line 17: What is the reason of introducing `FinalityProviderState` interface? On top of my head I cannot think of any other implementation of `FinalityProviderState`

finality-provider/service/fp_test_helper.go
SebastianElvis, line 1: Since this file contains all test utilities, do you think we should move it to `testutil/`?

finality-provider/service/finality_submitter.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	// update fp status according to the power for the last block
+	if hasPower && ds.state.GetStatus() != proto.FinalityProviderStatus_ACTIVE {
+		ds.mustSetStatus(proto.FinalityProviderStatus_ACTIVE)
+	}
+
+	if !hasPower && ds.state.GetStatus() == proto.FinalityProviderStatus_ACTIVE {
+		ds.mustSetStatus(proto.FinalityProviderStatus_INACTIVE)
+	}
```
SebastianElvis, line 153: I'm not sure if this function is supposed to modify FP's state in DB. This operation seems to belong to certain goroutines. Wdyt?

finality-provider/service/app.go
Lazar955: We do need to provide all these to the fpIns during it's instantiation. And we can't instantiate fpIns in the ctor of the fpApp, as it might not exist yet (fp not registered yet)

types/expected_finality_submitter.go
```diff
@@ -0,0 +1,22 @@
...
+	"context"
+)
+
+// FinalitySignatureSubmitter defines the interface for processing and submitting finality signatures
+// The implementation handles all retry logic internally
+type FinalitySignatureSubmitter interface {
+	// FilterBlocksForVoting processes a batch of blocks and returns ones that need voting
+	// It checks voting power, updates finality provider status, and filters out blocks
+	// that are below the last voted height or don't have voting power
+	FilterBlocksForVoting(ctx context.Context, blocks []BlockDescription) ([]BlockDescription, error)
```
KonradStaniec, line 13: I wonder whether `FilterBlocksForVoting` and `SubmitBatchFinalitySignatures` shouldn't be combined i.e remove `FilterBlocksForVoting` and implement its functionality privately in `SubmitBatchFinalitySignatures`

If you see how they are used:
```go
	processedBlocks, err := fp.finalitySubmitter.FilterBlocksForVoting(ctx, pollerBlocks)
	if err != nil {
		fp.reportCriticalErr(err)

		return
	}

	if len(processedBlocks) == 0 {
		return
	}

	res, err := fp.finalitySubmitter.SubmitBatchFinalitySignatures(ctx, processedBlocks)
```

it seems that caller is expected to first call `FilterBlocksForVoting` then `SubmitBatchFinalitySignatures`.  And there is nothing preventing him from just calling `SubmitBatchFinalitySignatures`

Of course then `SubmitBatchFinalitySignatures` would need to return more descriptive type that would indicate which blocks were submitted etc.

PR #547, opened 2025-07-22
api: consolidate `QueryActivatedHeight` and `QueryFinalityActivatationHeight` APIs

### LINKED ISSUE #529: rollup controller: implement QueryActivatedHeight

### LINKED ISSUE #545: consolidate `QueryActivatedHeight` and `QueryFinalityActivatationHeight`
There are two interfaces `QueryActivatedHeight` and `QueryFinalityActivatationHeight`. The former is for BTC staking activation height, and the latter is for BTC finality activation height.

For Babylon or any BSN, BTC finality activation height >= BTC staking activation height, and FP can vote only since BTC finality activation height. 

Having QueryFinalityActivationBlockHeight is good enough

SebastianElvis:
Resolves #545 
Resolves #529


SebastianElvis:
> Unrelated, what is the point of spec-consumer file, cause we will have docs on how to use the SDK. Is this like an intro or lighter version

It is a spec for the `ConsumerController` interface. We wrote this spec because BSNs might need this spec for developing their own BSNs

Issue #528, opened 2025-07-17
e2e: more e2e tests for rollup BSN FPs


SebastianElvis:
- [ ] make sure FP works with latest contract
- [ ] make sure CLIs work properly

Issue #521, opened 2025-07-14
Confirm if err strings in SubmitBatchFinalitySig match with those in rollup contract


gusin13:
seems like currently [these](https://github.com/babylonlabs-io/finality-provider/blob/21b4abc0132742cda12b78f187f58433c7835a76/bsn/rollup-finality-provider/clientcontroller/consumer.go#L228-L230) strings are aligned with babylon errs but not rollup contract errs

Issue #535, opened 2025-07-17
rollup: make rollup FP aware of the security features in the contract


SebastianElvis:
- [ ] Timestamped pub rand commit
- [ ] FP allowlist
- [x] Activation height + interval for finality signatures
- [x] ~~Rate limiting (this might not need any extra treatment in the rollup FP)~~ Any honest FP should not go beyond rate limit

SebastianElvis:
> FP allowlist – if I understand correctly, this is only needed for e2e tests, right? If so, it's already covered in the same PR above.

I think we also need to check whether the FP is in the allowlist somewhere in FP. I see two options:

1. In `runStartCmd`, check if the FP is in the allowlist, if no then exit
2. In `QueryFinalityProviderHasPower`, check if the FP is in the allowlist, if no then return 0

Currently I prefer option 1 because 

- If the FP is not allowlisted then there is no point for starting the finality provider
- 2 does not prevent FP from committing pub rand

Wdyt?

> Question – In the real-world scenario, who is responsible for adding an FP to the allowlist? Since we’re not enabling the allowlist during contract instantiation, I assume the FP operator would need to manually add themselves later by submitting a WASM transaction. Is that correct? Essentially, the same process as in the local deployment [here](https://github.com/babylonlabs-io/babylon-bsn-integration-deployment/blob/5480fcd39ae775eddd452491bea309f0654df111/deployments/rollup-bsn-demo/demo.sh#L149)?

It will be the responsibility of the contract admin, e.g., the entity/foundation of BSN, or security council

SebastianElvis:
> hmm on second thought about the FP allow list in `runStartCmd` this seems correct place, but in tests mostly we start the app instance directly using [instance.Start()](https://github.com/babylonlabs-io/finality-provider/blob/dfb6d3b1a557d04df00f914342a1a90acf03c495/finality-provider/service/fp_instance.go#L140)
> 
> maybe we should do in runStartCmd, QueryFinalityProviderHasPower and [rndCommitter.ShouldCommit](https://github.com/babylonlabs-io/finality-provider/blob/dfb6d3b1a557d04df00f914342a1a90acf03c495/finality-provider/service/fp_instance.go#L348) all 3 places, wdyt? 🤔

Yes make sense to me -- I think the FP allowlist check should be in all these places

Issue #527, opened 2025-07-17
e2e: refactor e2e test framework


SebastianElvis:
Currently, the e2e test framework is not versatile, and e2e tests for {Babylon, rollup, Cosmos} FPs have many duplicated code. It would be great to 

- make the `BaseTestManager` more versatile
- make the initialisation of containerised Babylon node consistent
- clean up `TestManager` for  {Babylon, rollup, Cosmos} FP e2e tests

PR #551, opened 2025-07-22
chore(rollup): add more tests for rollup bsn 

gusin13:
Closes 
https://github.com/babylonlabs-io/finality-provider/issues/521

Partially closes 
https://github.com/babylonlabs-io/finality-provider/issues/528
https://github.com/babylonlabs-io/finality-provider/issues/535

whats included in this pr
- adds e2e tests similar to babylon e2e
- rm individual unit style tests
- updates contract binary to latest and related config in test code
- sets up local anvil node using public image, block time set to 1s, finality sig interval set to 5 rollup blocks

whats not included in this pr/still pending
- ci related tests, there are 4-5 dedicated rollup-fpd commands which need to be tested, will add it in separate pr - [issue](https://github.com/babylonlabs-io/finality-provider/issues/559)
- some refactoring is needed to improve test managers - [issue](https://github.com/babylonlabs-io/finality-provider/issues/527)


finality-provider/service/finality_submitter.go
```diff
@@ -260,21 +261,27 @@ func (ds *DefaultFinalitySubmitter) submitBatchFinalitySignaturesOnce(ctx contex
...
 
-	// get proof list
-	proofBytesList, err := ds.proofListGetterFunc(
-		blocks[0].GetHeight(),
-		uint64(numPubRand),
-	)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get public randomness inclusion proof list: %w\nplease recover the randomness proof from db", err)
+	for _, block := range blocks {
+		// Get public randomness for this specific height
```
gusin13, line 269: ℹ️  this is a imp change - requesting for careful review

in the main branch we fetch all the pub rand starting from a block as we assume we are submitting blocks consecutively ([code ref](https://github.com/babylonlabs-io/finality-provider/blob/32947fbe9f349291cc6dd6500ded5d7b9094d7cc/finality-provider/service/finality_submitter.go#L266)) the problem in this approach is i got the following err
basically as we have defined a finality sig interval now, the inclusion proof didn't match the height at which the sig was submitted

`DEBUG: Error failed to submit finality signature to the consumer chain: failed to send finality signature messages: failed to reliably send messages: rpc error: code = Unknown desc = rpc error: code = Unknown desc = failed to execute message; message index: 1: The inclusion proof for height 120 does not correspond to the given height (125): execute wasm contract failed [CosmWasm/wasmd@v0.55.1/x/wasm/keeper/keeper.go:444] with gas used: '433521': unknown request`

the solution i figured out was to prepare the pub rand list relevant to the sig height, but that means we have not to iterate all blocks prepare the `proofBytesList` so the corrext indices are used, this is less efficient than the code in main branch but idk if there is a better way

bsn/rollup/clientcontroller/consumer.go
```diff
@@ -13,8 +13,7 @@ import (
 	bbnclient "github.com/babylonlabs-io/babylon/v3/client/client"
 	bbntypes "github.com/babylonlabs-io/babylon/v3/types"
 	btcstakingtypes "github.com/babylonlabs-io/babylon/v3/x/btcstaking/types"
-	finalitytypes "github.com/babylonlabs-io/babylon/v3/x/finality/types"
-	rollupfpconfig "github.com/babylonlabs-io/finality-provider/bsn/rollup-finality-provider/config"
+	rollupfpconfig "github.com/babylonlabs-io/finality-provider/bsn/rollup/config"
```
gusin13, line 16: bugbot run

itest/test-manager/base_test_manager.go
```diff
@@ -593,7 +594,7 @@ func StartEotsManagers(
 	}
 
 	babylonFpCfg.EOTSManagerAddress = eotsConfigs[0].RPCListener
-	consumerFpCfg.EOTSManagerAddress = eotsConfigs[1].RPCListener
+	consumerFpCfg.EOTSManagerAddress = eotsConfigs[0].RPCListener
```
gusin13, line 597: cc @Lazar955 this is bit recent change on main branch, this was causing some issue in tests

in L598 below we run a single eots server, but in the list we could have multiple?

temporarily i am using the same eots server for babylon, rollup bsn lmk if there is a better way here?

finality-provider/service/finality_submitter.go
SebastianElvis: Hmm good catch. Also I just realised that the original implementation might have some issue at the boundary of a pub rand commit -- suppose some blocks are covered by the previous pub rand commit and the other is covered by the next one, then wouldn't generating proofs fail?

bsn/rollup/e2e/rollup_test_manager.go
```diff
@@ -321,11 +340,16 @@ func deployCwContract(t *testing.T, bbnClient *bbnclient.Client, ctx context.Con
...
-		"bsn_id":       rollupBSNID,
-		"min_pub_rand": 100,
+		"admin":                       bbnClient.MustGetAddr(),
+		"bsn_id":                      rollupBSNID,
+		"min_pub_rand":                100,
+		"rate_limiting_interval":      10000,                     // test value
+		"max_msgs_per_interval":       100,                       // test value
+		"bsn_activation_height":       0,                         // immediate activation for tests
+		"finality_signature_interval": finalitySignatureInterval, // allow signatures every block for tests
+		"allowed_finality_providers":  nil,                       // allow none by default
```
SebastianElvis, line 352: Nice :+1:

SebastianElvis (approved):
Great work :+1: That's quite a lot of things.

Would be great if @Lazar955 can also take a look

Lazar955 (approved):
Lgtm, fix comments I left. 

@SebastianElvis @gusin13 Do these additioanl test feel redundant compared to the main FP ones?

Lazar955:
> #554 this made everything run in parallel, not sure why the regression in this PR? @gusin13

I can fix this in next commit not to block merging of this PR, let's move forward but in future let's not kick the rock down the road :D

Lazar955:
> > Do these additioanl test feel redundant compared to the main FP ones?
> 
> I don't think they are redundant. At the end of the day these tests need to exist in every FP.
> 
> We might consider making the test suite more versatile such that writing e2e tests for BSN FPs becomes easier. We have a Github issue for tracking this #527

ah actually yeah you are correct regarding tests

finality-provider/service/finality_submitter.go
Lazar955: Wait before we merge this. Both `GetPubRandList` and `proofListGetterFunc` fetch items for range. Can you explain what is the difference getting the one by one vs range?

Lazar955: Ah nvm we can have gaps in the block ranges. One idea to optimize is to iterate over block range to calculate ranges, but can be different PR

itest/babylon/babylon_e2e_test.go
```diff
@@ -41,12 +41,15 @@ const (
...
+// TODO: t.Parallel() is commented out due to conflict in ports/addresses
+// https://github.com/babylonlabs-io/finality-provider/issues/519
+
 // TestFinalityProviderLifeCycle tests the whole life cycle of a finality-provider
 // creation -> registration -> randomness commitment ->
 // activation with BTC delegation and Covenant sig ->
 // vote submission -> block finalization
 func TestFinalityProviderLifeCycle(t *testing.T) {
-	t.Parallel()
+	//t.Parallel()
```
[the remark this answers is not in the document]
gusin13: https://github.com/babylonlabs-io/finality-provider/pull/551#issuecomment-3128285697

not sure how to fix this tbh, this err is very transient and comes only when t.parallel() is added

finality-provider/service/finality_submitter.go
gusin13: hm @SebastianElvis I was checking code for proofListGetterFunc which uses [GetPubRandProofList](https://github.com/babylonlabs-io/finality-provider/blob/32947fbe9f349291cc6dd6500ded5d7b9094d7cc/finality-provider/store/pub_rand.go#L148C29-L148C48)

the db key seems to be chainId/Pk/height

now take this example1

Heights 98-100 belong to Commit A
Heights 101-102 belong to Commit B
proofListGetterFunc(98, 5)  // Try to get proofs for heights 98,99,100,101,102

imo this would not fail? because the db case is per height?

i think it would fail in gaps though if fp program is down for sometime and it misses 101-102 
Heights 98-100 belong to Commit A
Heights 101-102 belong to Commit B (this fails due to downtime somehow)

then the proof fetching will fail [here](https://github.com/babylonlabs-io/finality-provider/blob/32947fbe9f349291cc6dd6500ded5d7b9094d7cc/finality-provider/store/pub_rand.go#L166)
there is already a todo related to this [here](https://github.com/babylonlabs-io/finality-provider/blob/32947fbe9f349291cc6dd6500ded5d7b9094d7cc/bsn/rollup-finality-provider/clientcontroller/consumer.go#L266)

is this the same case you mentioned or maybe i misunderstood?


@Lazar955 hm i think yes this would be a performance concern, good idea if we have separate submitter for rollup case, in all other case (cosmos/babylon) they would use the default efficient one.
is it ok if i add a todo and a ticket for now for this and later we do this in separate pr. 
also about performance tracking - whats a good way to benchmark this - maybe we can do this as part of the new task as well, wdyt?
made a new ticket here 
https://github.com/babylonlabs-io/finality-provider/issues/560

SebastianElvis: > imo this would not fail? because the db case is per height?
> i think it would fail in gaps though if fp program is down for sometime and it misses 101-102

Hmm I think you are right. The DB saves proof for every height rather than re-generating the proof. And if there is a gap then yes fetching the proof will fail.

I don't find this a big performance issue though. The frequency of such operation is at most once per Babylon block interval, i.e., once per 10 seconds. Even after we reduce the block interval to sth like 2 seconds the DB access frequency seems fine.

gusin13:
> > > > #554 this made everything run in parallel, not sure why the regression in this PR? @gusin13
> > > 
> > > 
> > > I can fix this in next commit not to block merging of this PR, let's move forward but in future let's not kick the rock down the road :D
> > 
> > 
> > @Lazar955 I am not sure how to fix this, whenever I add t.Parallel() I keep getting these errors - this needs to be fixed separately - I see such issue is already tracked here #519
> > Screenshot 2025-07-28 at 1 29 51 PM
> 
> but this is not due to port conflict error, but due to FP restart, and db file not open

hm maybe my bad, i thought its due to some port conflicts, i had increased the sleep timer in tests when fp instance restarted which fixed the db open issue

thanks for enabling the t.Parallel() again 👍

PR #568, opened 2025-07-31
fix: correct logic for ctx signing start

Lazar955:


KonradStaniec (approved):
fixed condition looks good 👍 it would be great to add some tests catching this  (can be done in subsequent prs though)

PR #570, opened 2025-07-31
chore(rollup): don't allow FP program to start if not in allowlist

gusin13:
Don't allow FP program to start if FP not in allow list

~~For Babylon/Cosmos FP's the function returns always true
For Rollup FPs - it queries the rollup contract to find allow list~~

~~@SebastianElvis this is different from what we discussed [here](https://github.com/babylonlabs-io/finality-provider/issues/535#issuecomment-3134427979)
while coding I realized if we block the fp in fpinstance.Start() then we don't need to add this check in multiple places, as .Start() function is called from `runStartCmd` , also pub rand and finality sig submissions are started from this function~~


finality-provider/service/fp_instance.go
```diff
@@ -138,10 +139,19 @@ func newFinalityProviderInstanceFromStore(
 }
 
 func (fp *FinalityProviderInstance) Start(ctx context.Context) error {
-	if fp.isStarted.Swap(true) {
+	if fp.isStarted.Load() {
```
gusin13, line 142: would request reviewers to double check if this is correct cc @Lazar955 @SebastianElvis 

imo this seems to be bug in existing code, `fp.isStarted.Swap(true)` marks fp instance started, even if later point there is some err in `fpInstance.Start()` (lets say it fails at L169)

we should only mark the bool true once all validations are done and setup complete.

finality-provider/service/fp_instance.go
```diff
@@ -161,6 +171,11 @@ func (fp *FinalityProviderInstance) Start(ctx context.Context) error {
 
 	fp.quit = make(chan struct{})
 
+	// Mark as started only after ALL validation and setup is complete
+	if fp.isStarted.Swap(true) {
+		return fmt.Errorf("the finality-provider instance %s is already started", fp.GetBtcPkHex())
+	}
```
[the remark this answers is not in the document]
gusin13: I had moved the fp.IsStarted.Swap() to end b/c a unit test I added was calling .Start() multiple times, have removed the test now also fixed this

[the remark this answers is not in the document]
gusin13: @gitferry @Lazar955 @SebastianElvis can you explain why we have this `if fp.isStarted.Swap(true)`

1. the problem is if there is any err in Start(), we still mark the fp as running, shoudn't we mark the bool true only at end?
2. also what go routine this is referring to, isn't fp instance only started once?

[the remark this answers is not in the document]
gusin13: the ci test is failing b/c of this actually 
1. In the test i start(), which errs b/c fp was not in allowlist but the atomic bool is still marked as True
2. i add the fp in allowlist and start() again, this time it says "fp instance is already started" [ref](https://github.com/babylonlabs-io/finality-provider/actions/runs/16660732631/job/47157023166?pr=570#step:4:280)

clientcontroller/api/interface.go
```diff
@@ -112,6 +112,9 @@ type FinalityOperator interface {
 
 	// QueryFinalityProviderHighestVotedHeight queries the highest voted height of the given finality provider
 	QueryFinalityProviderHighestVotedHeight(ctx context.Context, fpPk *btcec.PublicKey) (uint64, error)
+
+	// QueryFinalityProviderInAllowlist queries whether the finality provider is in the allowlist
+	QueryFinalityProviderInAllowlist(ctx context.Context, fpPk *btcec.PublicKey) (bool, error)
```
SebastianElvis, line 117: Hmm since the allowlist feature will only likely to exist in rollup BSNs, I'm not sure if it's a good idea to make it mandatory for all consumer controllers.

What do you think we do this check in the `NewRollupBSNController` function, i.e., when constructing the rollup BSN consumer controller? Then

- The allowlist check is entirely within rollup BSN FPs
- We still achieve the goal of blocking non-allowlisted BSN FPs upon starting
- We won't have the race condition problem that we are facing right now

Lazar955: This seems like a good idea

gusin13: I agree tho the approach I used is causing interface pollution as this allowlist thing is not used by Babylon/Cosmos FPs

But I am still bit confused, we can't really add this check inside `NewRollupBSNController`, to perform the wasm query(QueryFinalityProviderInAllowlist ) the constructor needs to be initialized first, investigating again whats other available option

gusin13: ok I have implemented Option 2 now, added a check in `runStartCmd`, ideally we should add it in 

1. runStartCmd
2. QueryFinalityProviderHasPower
3. rndCommitter.ShouldCommit 

3 is not possible as this is a generic func used by all FPs (rollup, babylon, cosmos) so i also skipped 2
only did 1

ps implemented a hacky test case

bsn/rollup/cmd/rollup-fpd/daemon/start.go
```diff
@@ -75,6 +80,10 @@ func runStartCmd(ctx client.Context, cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("failed to create finality-provider app: %w", err)
 	}
 
+	if err := validateRollupFP(cmd.Context(), fpApp, fpStr, logger); err != nil {
+		return fmt.Errorf("failed to validate rollup finality provider: %w", err)
+	}
```
SebastianElvis, line 85: Good idea!

bsn/rollup/cmd/rollup-fpd/daemon/start.go
```diff
@@ -87,3 +96,43 @@ func runStartCmd(ctx client.Context, cmd *cobra.Command, _ []string) error {
...
+
+func validateRollupFP(ctx context.Context, fpApp *service.FinalityProviderApp, fpStr string, logger *zap.Logger) error {
+	rollupController, ok := fpApp.GetConsumerController().(*rollupfpcc.RollupBSNController)
+	if !ok {
+		return fmt.Errorf("expected RollupBSNController but got different controller type")
+	}
+
+	var fpToValidate *bbntypes.BIP340PubKey
+
+	if fpStr != "" {
```
Lazar955, line 108: nit: prefer fpStr == "" -> return err to avoid nested if, just better readability

Lazar955 (approved):
nice work! Much cleaner than prev solution

bsn/rollup/clientcontroller/msg.go
```diff
@@ -36,10 +36,11 @@ type SubmitFinalitySignatureResponse struct {
...
 type QueryMsg struct {
-	Config             *ContractConfig          `json:"config,omitempty"`
-	FirstPubRandCommit *PubRandCommit           `json:"first_pub_rand_commit,omitempty"`
-	LastPubRandCommit  *PubRandCommit           `json:"last_pub_rand_commit,omitempty"`
-	HighestVotedHeight *HighestVotedHeightQuery `json:"highest_voted_height,omitempty"`
+	Config                   *ContractConfig          `json:"config,omitempty"`
+	FirstPubRandCommit       *PubRandCommit           `json:"first_pub_rand_commit,omitempty"`
+	LastPubRandCommit        *PubRandCommit           `json:"last_pub_rand_commit,omitempty"`
+	HighestVotedHeight       *HighestVotedHeightQuery `json:"highest_voted_height,omitempty"`
+	AllowedFinalityProviders *struct{}                `json:"allowed_finality_providers,omitempty"`
```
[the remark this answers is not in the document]
gusin13: this is the query struct so it needs to be empty struct

the response struct is []string, defined [here](https://github.com/babylonlabs-io/finality-provider/pull/570/files#diff-0a8e07db552083ec916687616c54f9bb11a2e3486fc3b1845e5c309a6d0658b8R79)

PR #572, opened 2025-08-03
chore(rollup): ensure pub randomness is timestamped

gusin13:


types/pub_rand_commit.go
```diff
@@ -9,9 +9,10 @@ import (
...
 
 type PubRandCommit struct {
-	StartHeight uint64 `json:"start_height"`
-	NumPubRand  uint64 `json:"num_pub_rand"`
-	Commitment  []byte `json:"commitment"`
+	StartHeight  uint64 `json:"start_height"`
+	NumPubRand   uint64 `json:"num_pub_rand"`
+	Commitment   []byte `json:"commitment"`
+	BabylonEpoch uint64 `json:"babylon_epoch"`
 }
```
SebastianElvis, line 16: This is the struct shared across {Babylon, Cosmos, rollup} FPs so it might not be a good idea to have `BabylonEpoch` (which is specific to rollup) here.

Instead, wdyt we define another `PubRandCommit` struct under `bsn/rollup`? Then this struct is only used for Babylon FPs

bsn/rollup/clientcontroller/consumer.go
```diff
@@ -322,13 +332,73 @@ func (cc *RollupBSNController) QueryFinalityProviderHasPower(
...
 		nextKey = resp.Pagination.NextKey
 	}
+
+	return false, nil
+}
+
+// hasValidTimestampedPubRandomness validates that the FP has public randomness that:
+// 1. Exists (has committed public randomness that covers the specific height)
+// 2. Is Bitcoin timestamped (finalized)
+func (cc *RollupBSNController) hasValidTimestampedPubRandomness(ctx context.Context, fpPk *btcec.PublicKey, blockHeight uint64) bool {
```
SebastianElvis, line 342:
```suggestion
func (cc *RollupBSNController) hasTimestampedPubRandomness(ctx context.Context, fpPk *btcec.PublicKey, blockHeight uint64) bool {
```

SebastianElvis (commented):
LGTM, except for one comment on the PubRandCommit struct

types/pub_rand_commit.go
gusin13: ok so i made a different struct specific for rollup case, the new query i added is using this new struct now

the new query we added is not actually part of the interface but to properly fix the problem we also need to fix [this](https://github.com/babylonlabs-io/finality-provider/blob/ecf187708c18a16a15fb8bf5d31e137a33892061/clientcontroller/api/interface.go#L73) interface, currently (Babylon, Cosmos, Rollup) are all using a common struct, ideally this should be interface
this change will touch a lot of files, so I avoided to do this bigger change - maybe better if its part of separate pr 
I added a todo for now [ref](https://github.com/babylonlabs-io/finality-provider/pull/572/files#diff-6b02a1bba8ebe8e972db126bdde19fc6cc822b328b413d16ef2bb88567155f22R14)

SebastianElvis (approved):
LGTM! Need to be merged after https://github.com/babylonlabs-io/finality-provider/pull/576 though

PR #576, opened 2025-08-04
chore: fix PubRandCommit interface

gusin13:


types/pub_rand_commit.go
```diff
@@ -1,35 +1,17 @@
...
-// EndHeight returns the last height covered by this public randomness commitment
-func (prc *PubRandCommit) EndHeight() uint64 {
-	return prc.StartHeight + prc.NumPubRand - 1
+// PubRandCommit interface abstracts source-specific structs
+// (Babylon/Cosmos/Rollup FPs) to handle different JSON field names across
+// sources.
+type PubRandCommit interface {
+	EndHeight() uint64
+	Validate() error
 }
```
SebastianElvis, line 16: We need to define getter functions of common fields in the interface, including StartHeight, NumPubRand, and Commitment.

SebastianElvis (commented):
Also @Lazar955 wdyt on this change? The reason is that each type of FP has slightly different definitions of pub rand commit, due to the differences in BTC timestamping.

Lazar955 (commented):
Seems like right direction. The only nit is about struct naming suggestion:
Interface -> PubRandCommitter or PubRandCommitContract
structs -> PubRandCommit, we don't need to have prefix as they are defined in their own packages (avoid norevive lint)

PR #606, opened 2025-08-11
chore: make cosmos e2e run in parallel

### LINKED ISSUE #599: [Cosmos] Parallel e2e
Try to make cosmos e2e run in parallel. There is a huge problem as we need to call `SetAddressPrefixes` which invokes a global singleton `sdk.GetConfig()`. Try to use a global mutex in tests, altenatevly in makefile a script which calls go test to each invidiual e2e test so they run in a different process

Lazar955:
Cosmos e2e tests take way too long to run sequentially, now makes them parallel

closes #599

### DIFF
--- a/bsn/cosmos/e2e/bcd_consumer_e2e_test.go
+++ b/bsn/cosmos/e2e/bcd_consumer_e2e_test.go
@@ -1,3 +1,5 @@
+//go:build e2e_bcd
+
 package e2etest_bcd
 
 import (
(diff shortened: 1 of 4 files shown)


bsn/cosmos/e2e/bcd_consumer_e2e_test.go
```diff
@@ -42,7 +41,8 @@ import (
 // has started. This order is critical because the pub randomness loop takes time to initialize,
 // and without it, blocks won't get finalized properly.
 func TestConsumerFpLifecycle(t *testing.T) {
-	appparams.SetAddressPrefixes()
+	t.Parallel()
+	setBbnAddressPrefixesSafely()
```
SebastianElvis, line 45: Nice!

SebastianElvis (approved):

PR #636, opened 2025-08-18
chore: add mergify

Lazar955:


KonradStaniec (approved):

.mergify.yml
```diff
@@ -0,0 +1,25 @@
...
+  - name: default
+    queue_conditions:
+      - "#approved-reviews-by>1"
+      - base=main
+      - label=A:automerge
+    merge_conditions:
+      - "#approved-reviews-by>1"
+    commit_message_template: |
+      {{ title }} (#{{ number }})
+      {{ body }}
```
0xAleksaOpacic, line 11: Do we need body? As far as i know body is PR description, and sometimes someone would put quite big description + images...

0xAleksaOpacic (approved):
Looks good to me! We should share this to internal slack so other know how to use it. Also we need to keep in mind to adapt Backport rule when time comes

PR #647, opened 2025-08-22
chore: unify methods in eotsmanager

### LINKED ISSUE #648: Unify eotsmanager randomness pair list with and without interval methods

Lazar955:
closes #648


Lazar955:
https://github.com/Mergifyio backport release/v2.x

finality-provider/proto/finality_providers.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.32.0
+// 	protoc-gen-go v1.31.0
```
SebastianElvis, line 3: Why does this get downgraded?

eotsmanager/proto/eotsmanager.proto
```diff
@@ -73,6 +69,9 @@ message CreateRandomnessPairListRequest {
   uint64 start_height = 3;
   // num is the number of randomness pair list
   uint32 num = 4;
+  // interval is the optional height interval between consecutive randomness pairs
+  // If not set or 0, consecutive heights are used
+  optional uint64 interval = 5;
```
SebastianElvis, line 74: What do you think we have this as mandatory and require the client to use `1` explicitly for consecutive heights? 

Right now `interval == nil` and `interval == &1` have the same meaning, and type `*uint64` is a bit weird in Go

eotsmanager/localmanager.go
```diff
@@ -197,33 +197,23 @@ func LoadBIP340PubKeyFromKeyName(kr keyring.Keyring, keyName string) (*bbntypes.
...
-		// KEY DIFFERENCE: height increments by interval, not 1
-		height := startHeight + uint64(i)*interval // 100, 105, 110, 115...
+		var height uint64
+		if cfg.Interval != nil {
+			// Use interval: startHeight + i*interval
+			height = startHeight + uint64(i)*(*cfg.Interval)
+		} else {
+			// Consecutive heights: startHeight + i
+			height = startHeight + uint64(i)
+		}
```
SebastianElvis, line 216: Have we checked if `*cfg.Interval > 0`?

finality-provider/proto/finality_providers.pb.go
Lazar955: updated

PR #677, opened 2025-09-03
DNM: hacky bootstrap from last finalized height *DNM*

Lazar955:


Lazar955:
should have branched from rc3

PR #738, opened 2025-11-06
chore: prevent EOTS private key extraction via duplicate height validation

Lazar955:


eotsmanager/service/rpcserver.go
```diff
@@ -113,6 +114,17 @@ func (r *rpcServer) SignSchnorrSig(_ context.Context, req *proto.SignSchnorrSigR
 // SignBatchEOTS signs multiple EOTS in batch
 func (r *rpcServer) SignBatchEOTS(_ context.Context, req *proto.SignBatchEOTSRequest) (
 	*proto.SignBatchEOTSResponse, error) {
+	// Validate no duplicate heights before forwarding to manager (defense-in-depth)
+	heights := make([]uint64, len(req.SignRequests))
+	for i, signReq := range req.SignRequests {
+		heights[i] = signReq.Height
+	}
+
+	if err := util.ValidateNoDuplicateHeights(heights); err != nil {
```
RafilxTenfen: if we ask eots to sign the same height we just return the already created signature, we don't create another signature. But I guess it is okay to have additional validation

PR #777, opened 2026-02-23
fix: align pubrand and proof indexing with actual block heights

Lazar955:
- Fix height misalignment between public randomness/Merkle proofs and finality signatures when `FilterBlocksForVoting` produces non-consecutive block arrays (e.g. `[100, 102, 103]` after removing blocks
  where the FP lacks voting power)
  - Fetch pubrand and proofs for the full height span (`endHeight - startHeight + 1`) and index by `block.GetHeight() - startHeight` offset instead of array index
  - Apply the same fix to the test helper `SubmitBatchFinalitySignaturesAndExtractPrivKey` for consistency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+### Bug Fixes
+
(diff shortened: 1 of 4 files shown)


gitferry (approved):
Lgtm except for ci failure

